PR #6, opened 2019-12-12
Implement remote snapshotter as a proxy plugin

ktock:
This commit enables us to use remote snapshotter as a [containerd proxy plugin](https://github.com/containerd/containerd/blob/3e657de3afd98ecf6da7fe3ab7b1a1345305d8a9/PLUGINS.md#proxy-plugins). So we can bundle remote snapshotter and filesystem plugins in one container and plug it to containerd.

For more detail about containerization, see [docker-compose.yml](https://github.com/ktock/remote-snapshotter/blob/8840d1cca59dcf9b30a7aac6c1c6de4b26b57bea/script/make_wrapper/docker-compose-integration.yml.sh) and [this docker context](https://github.com/ktock/remote-snapshotter/tree/8840d1cca59dcf9b30a7aac6c1c6de4b26b57bea/script/make_wrapper/rs) .

We've already been able to specify the priority of filesystem plugins since https://github.com/ktock/remote-snapshotter/commit/26f211bb97be302016d0c31fff2ca6ff670e8800, so all issues mentioned in https://github.com/containerd/containerd/issues/3731#issuecomment-561623407 have solved.

cc: @jblomer @siscia


script/make_wrapper/containerd/config.toml
```diff
@@ -0,0 +1,4 @@
+[proxy_plugins]
+  [proxy_plugins.remote]
+    type = "snapshot"
+    address = "/var/lib/rs/rs.sock"
```
AkihiroSuda: nit: socks are typically located under /run

ktock: Thank you for your review. I'll fix it on the next commit.

PR #11, opened 2020-01-04
optimize: $0 -> os.Args[0]

AkihiroSuda:
Signed-off-by: Akihiro Suda <akihiro.suda.cz@hco.ntt.co.jp>


ktock:
@AkihiroSuda Thanks a lot!

PR #17, opened 2020-01-06
error out when no filesystem plugin is installed

AkihiroSuda:
Signed-off-by: Akihiro Suda <akihiro.suda.cz@hco.ntt.co.jp>


ktock:
Thanks a lot!

PR #35, opened 2020-01-28
Check layers with reasonable interval

ktock:


stargz/fs.go
```diff
@@ -115,6 +117,11 @@ func NewFilesystem(root string, config *Config) (snbase.FileSystem, error) {
 	if fs.httpCacheChunkSize == 0 {
 		fs.httpCacheChunkSize = defaultHTTPCacheChunkSize
 	}
+	interval := config.LayerValidInterval
+	if interval == 0 {
```
AkihiroSuda: check negative?

ktock: Thanks for the review. I added the description of the definition for this part.

Issue #30, opened 2020-01-20
support fetching the whole image in background after starting up the container


AkihiroSuda:
Fetching the whole image in background after starting up the container would make the container tolerant of registry failures.

ktock:
@AkihiroSuda #42 is addressing this issue.

ktock:
Solved in #42.

Issue #29, opened 2020-01-20
support prefetch for unoptimized images


AkihiroSuda:
The currently implementation supports prefetch only for `optimize`-d images that contain the landmark file: https://github.com/ktock/remote-snapshotter/blob/26c23b14959a2c05eaae004fb014243af72f8131/cmd/ctr-remote/sorter/sorter.go#L43-L57

We should allow prefetching for unoptimized images as well.
The amount of the prefetching should be configurable via the image handler wrapper.

PR #42, opened 2020-02-05
Fetch whole layer in background

ktock:
Background fetching disturbs more important "frontend" tasks(=authentication, redirection, reading files, etc...) so this commit includes a small task manager for less important "preemptive" tasks.

This is related to #30 and possibly give us performance improvement for unoptimized images(mentioned in #29).

### DIFF
--- a/stargz/fs.go
+++ b/stargz/fs.go
@@ -31,6 +31,7 @@
 package stargz
 
 import (
+	"bufio"
 	"bytes"
 	"context"
 	"encoding/json"
@@ -60,6 +61,7 @@ import (
 	"github.com/ktock/stargz-snapshotter/cache"
 	snbase "github.com/ktock/stargz-snapshotter/snapshot"
 	"github.com/ktock/stargz-snapshotter/stargz/handler"
+	"github.com/ktock/stargz-snapshotter/task"
 	"golang.org/x/sys/unix"
 )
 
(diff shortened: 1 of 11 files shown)


stargz/fs.go
```diff
@@ -157,7 +162,100 @@ type connection struct {
...
+		preemptiveSem:          semaphore.NewWeighted(concurrency),
+		privilegedTaskInterval: interval,
+	}
+}
+
+// preemptiveTaskManager manages privileged and preemptive task execution.
+// Preemptive tasks are some tasks aren't important and we don't want them
+// to use compute resources(CPU, NW, etc...) during more important tasks(=
+// privileged tasks) running.
+type preemptiveTaskManager struct {
```
AkihiroSuda: split go pkg?

AkihiroSuda: s/privileged/prioritized/ ?

stargz/fs.go
```diff
@@ -157,7 +162,100 @@ type connection struct {
...
+// invokePreemptiveTask invokes a preemptive task. The task is started only when
+// no privileged tasks are running. Privileged task's execution stops the
+// execution of all preemptive tasks. Priemptive task must be immutable and
+// resumable because this function suddenly stop and restart it.
+func (ts *preemptiveTaskManager) invokePreemptiveTask(do func(context.Context), timeout time.Duration) {
+	privilegedTaskAdded := func() <-chan int64 {
+		ch := make(chan int64)
+		go func() {
+			for ts.privilegedTasks <= 0 {
+				time.Sleep(10 * time.Millisecond)
```
AkihiroSuda: Use `sync.Cond`?

stargz/fs.go
```diff
@@ -157,7 +162,100 @@ type connection struct {
...
+		}()
+		return ch
+	}
+
+try:
+	// if nobody add privileged tasks in specified interval, this is a
+	// chance for us to do some preemptive tasks
+	select {
+	case <-privilegedTaskAdded():
+		time.Sleep(time.Second)
```
AkihiroSuda: ditto

stargz/fs.go
ktock: I'll fix it on the next commit.

stargz/fs.go
ktock: I'll fix it on the next commit.

stargz/fs.go
ktock: We need timeout here so we use channel instead of cond. There aren't easy ways to use cond with timeout?

task/task.go
```diff
@@ -0,0 +1,117 @@
...
+
+// DoPrioritizedTask tells the manager that we've done a prioritized task.
+func (ts *PreemptiveTaskManager) DonePrioritizedTask() {
+	atomic.AddInt64(&ts.prioritizedTasks, -1)
+}
+
+// InvokePreemptiveTask invokes a preemptive task. The task is started only when
+// no prioritized tasks are running. Prioritized task's execution stops the
+// execution of all preemptive tasks. Priemptive task must be immutable and
+// resumable because the manager suddenly stops and restarts the task.
```
AkihiroSuda: Needs more godoc about how a task can be stopped and restarted?

ktock: Added more description about this manager.

PR #47, opened 2020-02-17
Prefetch layers in parallel

### LINKED ISSUE #37: Do not Mount() on Prepare(), or parallelize Prepare()
https://github.com/ktock/stargz-snapshotter/pull/33

`Prepare()` operations seems happening in sequential and limiting the performance.

If we skip `Mount()` on `Prepare()` and defer lazy-pull operations until actual getdents happens, probably we can see significant decrease in the startup latency.

Or maybe we should modify the unpacker so that it can unpack layers in parallel.

ktock:
This commit enables us to prefetch layers in parallel using goroutine and fixes #37.

To make sure that prefetch completes when we "actually" use these layers(e.g. unpacking layers on it or adding a writable layer on it for the container execution) we synchronize the prefetch completion using `filesystem.Check()` as following.

- When layers are actually used, containerd queries list of mount points(`mount.Mount[]`) and at this point, we check the layer status by invoking `filesystem.Check()` for each layer. 
- The first `filesystem.Check()` of the layer waits for the prefetch completion with a reasonable timeout(10 sec).

### DIFF
--- a/snapshot/snapshot.go
+++ b/snapshot/snapshot.go
@@ -196,7 +196,7 @@ func (o *snapshotter) Usage(ctx context.Context, key string) (snapshots.Usage, e
 }
 
 func (o *snapshotter) Prepare(ctx context.Context, key, parent string, opts ...snapshots.Opt) ([]mount.Mount, error) {
-	m, err := o.createSnapshot(ctx, snapshots.KindActive, key, parent, opts)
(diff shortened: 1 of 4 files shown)


ktock:
Seems improving well in terms of `pull` but the *first* `create` takes longer because it waits for prefetch completion.

PR #48, opened 2020-02-21
Hide state dir from ReadDir req to avoid affecting diffing result

### LINKED ISSUE #31: RFC: hide .remote-snapshotter directory from getdents()
Showing `.remote-snapshotter` directory in `getdents(2)` may affect `differ` implementations like https://github.com/containerd/containerd/blob/master/diff/walking/differ.go .

So maybe we should hide `.remote-snapshotter` from `getdents(2)`.

This is akin to hidden `.zfs` directories used in ZFS.

ktock:
Fixes #31

This commit hides state dir(`/.stargz-snapshotter`) from getdents(2) to avoid affecting differ implementations.

```
# ctr-remote run --rm -t --snapshotter=stargz registry2:5000/ubuntu:18.04 test /bin/bash
root@32d4520a7fb5:/# ls -a /
.  ..  bin  boot  dev  etc  home  lib  lib64  media  mnt  opt  proc  root  run  sbin  srv  sys  tmp  usr  var
root@32d4520a7fb5:/# ls -a /.stargz-snapshotter
.
..
sha256:e43418d927530df8736bfc362d26997f0c96bfcf90b6fb594510d54da0d3f366.json
sha256:e5a5544b4500edc7341910013633704a3edac54ee52990ac628a5fbc5a591075.json
sha256:f05c0f4ff792f589752764487e35ce1b60eaad399353b2fce046874b1b74df3b.json
sha256:feb2e5ef2735310a4116fbe6571cbcda2640f94ed5e2305e56d1b3b1c7995f6a.json
```

In the future, we should also make the state directory read-only(or prohibit to create files in) on the container's **merged** rootfs as well. This is because once we create a file in the state directory, overlayfs will create a directory which has the same name as the state directory in the upper layer directory and the state directory will be visible with getdents.

This PR includes only hiding the state dir and read-only state dir is out-of-scope for now.

### DIFF
--- a/stargz/fs.go
+++ b/stargz/fs.go
@@ -524,15 +524,6 @@ func (n *node) OpenDir(context *fuse.Context) ([]fuse.DirEntry, fuse.Status) {
 		}
 	}
 
-	// Append state directory in "/".
-	if n.e.Name == "" {
-		ents = append(ents, fuse.DirEntry{
-			Mode: syscall.S_IFDIR | n.s.mode(),
-			Name: stateDirName,
-			Ino:  n.s.ino(),
-		})
-	}
-
 	sort.Slice(ents, func(i, j int) bool { return ents[i].Name < ents[j].Name })
 	return ents, fuse.OK
 }
--- a/stargz/fs_test.go
+++ b/stargz/fs_test.go
@@ -535,7 +535,27 @@ func hasNodeXattrs(entry, name, value string) check {
 }
 
 func hasStateFile(id string) check {
+	isExist := func(name string, ents []fuse.DirEntry) bool {
+		for _, e := range ents {
+			if e.Name == name {
+				return true
+			}
+		}
+		return false
+	}
+
 	return func(t *testing.T, root *node) {
+
+		// Check the state dir is hidden on OpenDir for "/"
+		ents, status := root.OpenDir(nil)
+		if status != fuse.OK {
+			t.Errorf("failed to open root directory: %v", status)
+			return
+		}
+		if isExist(stateDirName, ents) {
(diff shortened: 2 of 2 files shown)


ktock:
`glassfish:4.1-jdk8` takes too much time for container creation. I'll take one more benchmark with empty push. But this commit doesn't seem to be the cause.

ktock:
Sometimes, `glassfish:4.1-jdk8` takes too much time for container creation. But this commit doesn't seem to be the cause. And also, there seems large dispersion for the result maybe because of the NW condition and/or instance location, etc. Let's work on it with another PR.

### 1st time

mode | pull(sec) | create(sec) | run(sec)
-- | -- | -- | --
legacy | 41.17897605895996 | 0.10247516632080078 | 3.1075639724731445
stargz | 11.927096128463745 | 0.11415910720825195 | 13.317368030548096
estargz | 22.953274965286255 | 21.620684146881104 | 6.718463897705078


### 2nd time

mode | pull(sec) | create(sec) | run(sec)
-- | -- | -- | --
legacy | 39.5100679397583 | 0.09084105491638184 | 2.75496506690979
stargz | 2.7749650478363037 | 0.14973998069763184 | 11.852347135543823
estargz | 17.710061073303223 | 11.861916065216064 | 5.867160081863403

### 3rd time

mode | pull(sec) | create(sec) | run(sec)
-- | -- | -- | --
legacy | 30.083626985549927 | 0.04444289207458496 | 2.4963369369506836
stargz | 2.3760488033294678 | 0.047811031341552734 | 12.571553945541382
estargz | 3.8805949687957764 | 5.056465148925781 | 3.3340909481048584

AkihiroSuda:
Make sure to document this

PR #52, opened 2020-03-04
[WIP] buildkit integration

ktock:
This is an experimental integration with buildkit for speeding up fetching base image.

Our patched version of buildkit is [here](https://github.com/ktock/buildkit/tree/remote-snapshotter).

This commit includes benchmark scripts to measure the time for buliding sample images.
See [this doc](https://github.com/ktock/stargz-snapshotter/blob/21a8e3c832e86b9ba137e6aa9ef7a6d02aef1eef/script/buildkit/testing.md) for manual testing.

Though it seems good for the lazy distribution of the base images, we currently have two problems for exporting output image.

1. Archiving layers takes a long time. It is because of the low READ performance including fetching contents from the registry.
2. We only support exporting type `tar`. It is because when we use remote snapshotters, containerd doesn't store the image contents in the content store but buildkit needs to use these contents for exporting images.

### DIFF
--- a/.github/workflows/buildkit.yml
+++ b/.github/workflows/buildkit.yml
@@ -1,0 +1,25 @@
+name: Buildkit
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+
+jobs:
+  buildkit-integration:
+    runs-on: ubuntu-18.04
+    name: BuildkitIntegration
+    env:
+      BUILDKIT_RESULT_DIR: ${{ github.workspace }}/buildkit/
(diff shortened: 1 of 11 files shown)


AkihiroSuda:
Can you open a proposal for BuildKit changes in https://github.com/moby/buildkit/issues

cc @tonistiigi

ktock:
Yes. Before that, please let me re-open this PR from my branch.

ktock:
Reopened on #53.

Issue #50, opened 2020-02-29
Add the list of the pre-converted images to README


AkihiroSuda:

AkihiroSuda:
Also better to have benchmark result on README.

@ktock WDYT?

ktock:
Yes, I agree that we need them. I'll work on it soon.

PR #55, opened 2020-03-06
Add y-label to benchmarking result histogram and fix the title

ktock:
We need enough information for understanding the histogram, aiming to add it to README (#50).


ktock:
Seems fine. Ready for review.

result

PR #56, opened 2020-03-06
Avoid duplicated authentications on prefetch and background-fetch

ktock:
Currently, we re-authenticating RoundTripper for each reading during prefetch and background-fetch and it costs extra time for each authentication.

https://github.com/containerd/stargz-snapshotter/blob/28ff0b6d5fe9b5a670d2ce46c2d059835033737b/stargz/fs.go#L252

https://github.com/containerd/stargz-snapshotter/blob/28ff0b6d5fe9b5a670d2ce46c2d059835033737b/stargz/fs.go#L280

This commit solves this by enabling sharing transports between prefetch and background-fetch.

### DIFF
--- a/stargz/fs.go
+++ b/stargz/fs.go
@@ -175,10 +175,6 @@ type filesystem struct {
 	debug                 bool
 }
 
-type readerAtFunc func([]byte, int64) (int, error)
-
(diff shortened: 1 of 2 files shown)


ktock:
Looks good. Ready for review.
result

PR #58, opened 2020-03-12
Add pre-converted images list and benchmarking result to docs

ktock:
Fixes:  #50

This commit includes
- Adding benchmarking result and fixing typos for README
- Adding a document about pre-converted images
- Adding detailed information about pre-converted images for the benchmarking
  script and fixing bugs related to it


ktock:
Ready for review :raised_hands:

README.md
```diff
@@ -1,13 +1,14 @@
...
+Pulling image is one of the time-consuming steps in the container lifecycle. Research shows that time for pulling accounts for 76% of container startup time[[FAST '16]](https://www.usenix.org/node/194431). *Remote snapshotter* is a solution discussed in containerd community. *Stargz Snapshotter* is an implementation of the remote snapshotter which aims to standard-compatible remote snapshots leveraging [stargz image format by CRFS](https://github.com/google/crfs).
 
-Related discussion of the snapshotter in containerd community:
-- [Support remote snapshotter to speed up image pulling#3731@containerd](https://github.com/containerd/containerd/issues/3731)
-- [Support `Prepare` for existing snapshots in Snapshotter interface#2968@containerd](https://github.com/containerd/containerd/issues/2968)
-- [remote filesystem snapshotter#2943@containerd](https://github.com/containerd/containerd/issues/2943)
+Related discussions of the snapshotter in containerd community are following,
```
AkihiroSuda: Do we still need these links?

README.md
```diff
@@ -25,11 +26,13 @@ user	0m0.112s
 sys	0m0.008s
 ```
 
+In the above example, we use a docker registry serving on `localhost`. For the real-world benchmarking result with Docker Hub, see [the later section](#benchmarking-result).
```
AkihiroSuda: The localhost result is kinda boring. Just show real-world results.

docs/pre-converted-images.md
```diff
@@ -0,0 +1,35 @@
...
+
+- `org`: Legacy image copied from `docker.io/library` without optimization. Layers are normal tarballs.
+- `sgz`: Stargz-formatted version of the `org` images. Layers are stargz archives and we can see performance improvement on the pull operation.
+- `esgz`: Heavily optimized version of the `org` images, based on the workload. `ctr-remote images optimize` command is used for the optimization. We can see the performance improvement on run operation as well as pull.
+
+`Optimized Workload` column describes workloads used for building `esgz` images. We optimized these images for benchmarking which is based on [HelloBench](https://github.com/Tintri/hello-bench) so we specified "hello-world"-like workloads for the command. See [benchmarking script](/script/benchmark/hello-bench/src/hello.py) for the exact command option specified for `ctr-remote images optimize`. 
+
+|Image Name|Optimized Workload|
+---|---
+|`docker.io/ktokunaga/alpine:3.10.2-{{ suffix }}`|Executing `echo hello` on the shell|
```
AkihiroSuda: I'd suggest creating a dedicated Docker Hub org. Can be follow-up in future.

README.md
```diff
@@ -1,35 +1,21 @@
...
-sys	0m0.008s
-```
+`legacy` shows the startup performance when we use containerd's default snapshotter (`overlayfs`) with images copied from `docker.io/library` without optimization. For this configuration, containerd pulls entire image contents and `pull` operation takes accordingly. When we use stargz snapshotter with `stargz` images we are seeing performance improvement on the `pull` operation because containerd can start the container before the entire image contents locally available but fetches each file on-demand. When we use the heavily optimized version of images(`estargz`) the performance of `run` operation (= running ["hello world"-like programs](./script/benchmark/hello-bench/src/hello.py) of each image) is much faster than `stargz`. This is because stargz snapshotter prefetches and caches some files which will be most likely accessed during container execution. Stargz snapshotter waits for the first container creation until the prefetch completes so `create` sometimes takes longer than other types of image. But this wait only occurs just after the pull completion until the prefetch completion and it's much shorter than waiting for downloading all files of all layers.
```
AkihiroSuda: Is `estargz` "much" faster than `stargz`?

ktock: I agree with `much faster` is misleading. I think `we can mitigate the performance drawback observed in stargz images` is the more precise expression with the following context:

> ... When we use stargz snapshotter with `stargz` images we are seeing performance improvement on the `pull` operation because containerd can start the container before the entire image contents locally available and fetches each file on-demand. But at the same time, we see the performance drawback for `run` operation because each access to files takes extra time for fetching them from the registry. When we use the further optimized version of images(`estargz`) we can mitigate the performance drawback observed in `stargz` images. ...

docs/pre-converted-images.md
ktock: I agree with this. Let's work on it on another thread.

README.md
AkihiroSuda: looks good

PR #59, opened 2020-03-16
Enable to cancel optimization on getting signal

ktock:
Currently, once image optimization starts we cannot cancel it and need to wait for the completion. This commit enables us to cancel the optimization.

### DIFF
--- a/cmd/ctr-remote/sampler/sampler.go
+++ b/cmd/ctr-remote/sampler/sampler.go
@@ -65,6 +65,7 @@ func Run(bundle string, config v1.Image, period int, opts ...Option) error {
 		syscall.SIGINT,
 		syscall.SIGTERM,
 		syscall.SIGQUIT)
+	defer signal.Stop(sc)
 
 	select {
 	case <-sc:


AkihiroSuda:
What does happen if we don't `cancel()` and just exit on SIGINT?

ktock:
I think just exiting on SIGINT is also fine and simpler. I reimplemented the patch.

AkihiroSuda:
If you want to handle SIGINT, you'd better gracefully shutdown with `cancel()`.
My question is that what was the previous behavior? I assumed it had exited on SIGINT by default.

ktock:
You are right. It seems missing `signal.Stop` corresponding to `signal.Notify` in `sampler` and this keeps relaying all signals instead of exiting the code.
https://github.com/containerd/stargz-snapshotter/blob/877fc5452a9c61adf314d7b010753688e86c996f/cmd/ctr-remote/sampler/sampler.go#L63-L67

ktock:
I reimplemented the patch to fix the bug.

Issue #61, opened 2020-03-20
Support mirror registries


ktock:
We should enable to configure mirror registries. There are two possible ways to achieve this.
- Makeing containerd pass the final result of resolved registry URL to this snapshotter.
- Enabling to configure mirror registries in stargz-snapshotter's config file. The mirror's management will be separated from containerd and image refs will be resolved in this snapshotter.

PR #62, opened 2020-03-25
Support mirror registries

ktock:
#61

This patch enables to configure mirror registries for this snapshotter, independent to containerd.
We can configure mirror registries in our config file like:

```toml
[resolver.mirrors]
registryexample.io = ["mirrorexample.com:5000", "mirrorexample2.com:5000"]
```
Authentication is done with `~/.docker/config.json` like as before.

The patch is a bit large but what is done is relatively simple.
- Moved resolver-related codes from `/stargz/fs.go` to `/stargz/remote/...`
- Added mirror registry support to `/stargz/remote/resolver.go` and added some tests
- Renamed `URLReaderAt`(`stargz/remote/urlreaderat.go`) to `Blob` (`stargz/remote/blob.go`) because its feature became richer by some codes (authentication of `RoundTripper` and getter of the blob size info) ported from `/stargz/fs.go`.

### DIFF
--- a/stargz/remote/resolver.go
+++ b/stargz/remote/resolver.go
@@ -1,0 +1,270 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+/*
+   Copyright 2019 The Go Authors. All rights reserved.
+   Use of this source code is governed by a BSD-style
+   license that can be found in the NOTICE.md file.
+*/
+
+package remote
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+
+	"github.com/containerd/containerd/reference/docker"
+	"github.com/containerd/stargz-snapshotter/cache"
+	"github.com/google/go-containerregistry/pkg/authn"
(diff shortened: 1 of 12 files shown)


ktock:
Ready for review:pray:

stargz/remote/resolver.go
```diff
@@ -0,0 +1,269 @@
...
+)
+
+const (
+	defaultChunkSize     = 50000
+	defaultValidInterval = 60 * time.Second
+)
+
+type ResolverConfig struct {
+	Insecure []string            `toml:"insecure"`
+	Mirrors  map[string][]string `toml:"mirrors"`
```
AkihiroSuda: `map[string]MirrorConfig` or just `[]MirrorConfig` might be better for extensibility

stargz/remote/resolver.go
```diff
@@ -0,0 +1,280 @@
...
+	)
+	named, err := docker.ParseDockerRef(ref)
+	if err != nil {
+		return nil, err
+	}
+	var hosts []registryHost
+	for _, m := range r.config[docker.Domain(named)].Mirrors {
+		hosts = append(hosts, registryHost{
+			insecure: strings.HasPrefix(m, httpPrefix),
+			host:     strings.TrimPrefix(m, httpPrefix),
```
AkihiroSuda: Don't you also need to trim `https://` prefix?

stargz/remote/resolver.go
```diff
@@ -0,0 +1,280 @@
...
+		return 0, fmt.Errorf("failed HEAD request with code %v", res.StatusCode)
+	}
+	return strconv.ParseInt(res.Header.Get("Content-Length"), 10, 64)
+}
+
+func authnTransport(ref name.Reference, tr http.RoundTripper) (http.RoundTripper, error) {
+	// Authn against the repository using `~/.docker/config.json`
+	auth, err := authn.DefaultKeychain.Resolve(ref.Context())
+	if err != nil {
+		return nil, fmt.Errorf("failed to resolve the reference %q: %v", ref, err)
```
AkihiroSuda: nit: use `errors.Wrap` whenever possible (in follow-up PR)

ktock:
As parsing hostname string seems a bit complex, I choosed to struct the config with `[]MirrorConfig` as you pointed. Sorry for fixing same place repeatedly:pray:

Issue #44, opened 2020-02-12
Build using `go mod`?


siscia:
Cheers guys!

I am trying to build our own package for cvmfs.

Being a new project, I believe the suggested way to build it would be through the `go mod`, however, the standard `go build` fails on 

```
# github.com/ktock/stargz-snapshotter/snapshot
../../../../pkg/mod/github.com/ktock/stargz-snapshotter@v0.0.0-20200212083925-a380be3bb5bc/snapshot/snapshot.go:330:52: undefined: snapshots.WalkFunc
../../../../pkg/mod/github.com/ktock/stargz-snapshotter@v0.0.0-20200212083925-a380be3bb5bc/snapshot/snapshot.go:336:25: too many arguments in call to storage.WalkInfo
```

After few experiments I can guess that the issue is the use of the `go mod`.

What is the policy of this project? Can/Should I use the standard `go mod` when using this project as dependency or I should find some other hack?

Can I help somehow?

siscia:
Aw I see!

https://github.com/containerd/containerd/issues/3031

ktock:
@siscia Is this solved? I think we should make this project `go mod` friendly in the near future.

ktock:
And thanks for trying this snapshotter out! :raised_hands:

siscia:
No unfortunately not.

I believe that the problem is the issue I posted above, as long as containerd is not `go mod` friendly, it would be quite hard for project that relies on containerd to be `go mod` friendly.

I briefly try yesterday to make containerd friendly to `go mod` without much success.

I want to finalize and test our snapshotter today and then I will try again :)

siscia:
Wondeful thanks guys!

PR #65, opened 2020-03-30
Move to go modules

ktock:
Fixes: #44 
We currently don't have so much dependencies so it seems good timing to move to go modules.
(BTW do we still need vendoring?)


ktock:
Ready for Review :pray:

AkihiroSuda:
> BTW do we still need vendoring?

Let's keep it for now

Issue #63, opened 2020-03-26
Enable to get credentials through Secrets on Kubernetes


ktock:
Kubernetes [allows users to pass registry creds (secrets) to runtimes](https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/) via CRI but by design stargz snapshotter doesn't share the creds with containerd.

One possible solution is to directly access k8s API and sync secrets to this snapshotter. Fortunately, we can use [k8s-based keychain](https://github.com/google/go-containerregistry/tree/35f57d7d493094506010dc4a52894a05d3be6c6f/pkg/authn/k8schain) supported by go-containerregistry which we are currently using for `~/.docker/config.json`-based credential management.

PR #66, opened 2020-04-01
Support kubeconfig-based keychain

ktock:
Fixes: #63
On kubernetes clusters, users can directly pass registry creds to containerd through CRI using secrets but by design, stargz-snapshtter doesn't share these creds.

This commit solves the issue by supporting kubeconfig-based keychain. Since this commit, stargz snapshotter syncs and uses "kubernetes.io/dockerconfigjson" secrets in the kubernetes cluster using specified kubeconfig. So stargz snapshotter can deal with images in private registries using kubernetes secrets. 

We can specify the path to kubeconfig in the config file like this:
```toml
[kubeconfig_keychain]
enable_keychain = true
kubeconfig_path = "/etc/kubernetes/snapshotter/config.conf"
```
When no path provided, snapshotter disables this synchronization.

This commit also adds a test for this feature using KinD. This test checks if stargz snapshotter can provide remote snapshots from a private registry (a docker container connected to KinD network) using secrets. Currently, this test is limited to this feature, but I think we can also use this Kuberenetes environment for other use-cases (e.g. k8s e2e tests).


Dockerfile
```diff
@@ -0,0 +1,43 @@
...
+
+# build customized version of containerd which supports remote snapshots via CRI
+FROM golang-base AS containerd-builder
+RUN apt-get update -y && \
+    apt-get install -y libbtrfs-dev libseccomp-dev && \
+    git clone https://github.com/ktock/containerd \
+              $GOPATH/src/github.com/containerd/containerd && \
+    cd $GOPATH/src/github.com/containerd/containerd && \
+    git checkout 7b2ddf1589694a52f5df5ff351bd2764fb1d0a3a && \
+    GO111MODULE=off make -j4 && DESTDIR=/out/ make install
```
AkihiroSuda, line 25: `Is -j4` meaningful?

stargz/keychain/kubeconfig.go
```diff
@@ -0,0 +1,150 @@
...
+			logCtx.Info("Syncing keychain with kubeconfig on the system")
+			if secrets, err := syncSecrets(ctx, kubeconfigPath); err == nil {
+				logCtx.Info("Completed to sync keychain")
+				kc.secretsMu.Lock()
+				kc.secrets = secrets
+				kc.secretsMu.Unlock()
+			} else {
+				logCtx.WithError(err).Infof("failed to sync secrets with kubeconfig")
+			}
+			time.Sleep(time.Duration(syncInterval) * time.Second)
```
AkihiroSuda, line 75: Use Watch API

AkihiroSuda: see https://engineering.bitnami.com/articles/a-deep-dive-into-kubernetes-controllers.html

ktock:
@AkihiroSuda Thanks for the feedback! Reimplemented keychain in controller's manner. PTAL

stargz/keychain/kubeconfig.go
```diff
@@ -0,0 +1,270 @@
...
+	return nil, fmt.Errorf("creds for %q not found", at.target.String())
+}
+
+func newKeychain(ctx context.Context, kubeconfigPath string) *keychain {
+	kc := &keychain{
+		config: make(map[string]*dcfile.ConfigFile),
+	}
+	logCtx := log.G(ctx).WithField("kubeconfig", kubeconfigPath)
+	go func() {
+		logCtx.Debugf("Waiting for kubeconfig being installed...")
```
AkihiroSuda, line 95: kubeconfig should be already installed, no?

stargz/keychain/kubeconfig.go
```diff
@@ -0,0 +1,270 @@
...
+// and when the file is actually provided, this keychain tries to access API
+// server using the file. This enables us to lazily provide kubeconfig file to
+// this snapshotter.
+// TODO: support update of kubeconfig file
+func NewKubeconfigKeychain(ctx context.Context, config KubeconfigKeychainConfig) authn.Keychain {
+	var (
+		kubeconfigPath = config.KubeconfigPath
+	)
+	if kubeconfigPath == "" { // If no path provided, disable syncing.
+		log.G(ctx).Warn("No kubeconfig provided; Disabling sync")
```
AkihiroSuda, line 65: No support for `$KUBECONFIG`?

stargz/keychain/kubeconfig.go
ktock: Yes, it should be. But I thought this waiting was convenient for the environments (e.g. single node cluster with containerized apiserver) where stargz snapshotter need to start before everything, including booting containerd/kubelet/apiserver and configuring users/roles. WDYT?

stargz/keychain/kubeconfig.go
ktock: Thanks for the comment. I added `KUBECONFIG` support.

stargz/keychain/kubeconfig.go
AkihiroSuda: LGTM, but add a document about that

ktock: Added description about this on README. We can merge it after CI being green.

PR #71, opened 2020-04-09
Refactor testing scripts

ktock:
`script/make.sh` was a centralized entrypoint for all tests under `script` directory. It was too large and hard to maintain because it was shared by all tests under `script` and contained glue codes of each test. This commit refactor it by moving per-test specific codes contained in `script/make.sh` into each test directories.

This commit also changed CRI Validation test to use *extended* stargz instead of the original stargz. This improves container's read performance during tests and hopefully mitigates recent test flaky (especially for performance-sensitive tests (e.g. [`ContainerPID` test](https://github.com/kubernetes-sigs/cri-tools/blob/fad4c4ea1c3a3208050e9f8e34fa4a38065f8af2/pkg/validate/security_context_linux.go#L218-L241) whose result is affected by the performance of nginx process).


ktock:
Ready for review.

PR #72, opened 2020-04-14
Add configuration for default prefetch size

ktock:
Fixes: #29

Currently, stargz snapshotter doesn't support prefetch for non-optimized stargz
images.

This commit adds a configuration value for the size to prefetch layers. We can
configure it through the config file and also can override it with snapshot
labels. Non-optimized images will be prefetched respecting the value. If
optimized images are provided, the optimized size to prefetch will be
used. During optimization, non-accessed layers are marked as "no prefetch". When
we pull these layers all the above values are disabled and no prefetch will
occur.

By default, no prefetch occur (prefetch size = 0) for non-optimized images.

TODO list:
* [x] Getting approved this patch
* [x] Updating optimized stargz images on `docker.io/stargz`
* [x] Re-running the benchmarking job
* [x] Getting merged this PR if CI is green

cc: @AkihiroSuda

### DIFF
--- a/cmd/ctr-remote/commands/rpull.go
+++ b/cmd/ctr-remote/commands/rpull.go
@@ -94,7 +94,7 @@ func pull(ctx context.Context, client *containerd.Client, ref string, config *co
 		containerd.WithSchema1Conversion,
 		containerd.WithPullUnpack,
 		containerd.WithPullSnapshotter(remoteSnapshotterName),
-		containerd.WithImageHandlerWrapper(stargz.AppendInfoHandlerWrapper(ref)),
+		containerd.WithImageHandlerWrapper(stargz.AppendInfoHandlerWrapper(ref, 10*1024*1024)),
 	}...); err != nil {
 		return err
 	}
--- a/stargz/reader/reader.go
+++ b/stargz/reader/reader.go
@@ -39,7 +39,13 @@ import (
 )
 
 const (
+	// PrefetchLandmark is a file entry which indicates the end position of
+	// prefetch in the stargz file.
 	PrefetchLandmark = ".prefetch.landmark"
+
+	// NoPrefetchLandmark is a file entry which indicates that no prefetch should
+	// occur in the stargz file.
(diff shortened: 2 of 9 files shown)


cmd/ctr-remote/commands/rpull.go
```diff
@@ -94,7 +94,7 @@ func pull(ctx context.Context, client *containerd.Client, ref string, config *co
 		containerd.WithSchema1Conversion,
 		containerd.WithPullUnpack,
 		containerd.WithPullSnapshotter(remoteSnapshotterName),
-		containerd.WithImageHandlerWrapper(stargz.AppendInfoHandlerWrapper(ref)),
+		containerd.WithImageHandlerWrapper(stargz.AppendInfoHandlerWrapper(ref, 10000000)),
```
AkihiroSuda, line 97: `10000000` -> `10 * 1000 * 1000` (or `10 * 1024 * 1024`?) for readability

stargz/reader/reader.go
```diff
@@ -39,7 +39,8 @@ import (
 )
 
 const (
-	PrefetchLandmark = ".prefetch.landmark"
+	PrefetchLandmark   = ".prefetch.landmark"
+	NoPrefetchLandmark = ".no.prefetch.landmark"
```
AkihiroSuda, line 43: Make sure to document these files

ktock: I'll add a doc about e-SGZ structure with another PR.

cmd/ctr-remote/commands/rpull.go
ktock: Thanks for the comment. Fixed it.

ktock:
something is wrong. I'll take a look at.
result

ktock:
Re-invoked benchmark without code modification (empty pushed). And the performance seems fine now. The previous benchmark was the first time for pulling newly updated eSGZ images so some performance overhead might have been added because of CDN-caching-related factors. We might need to investigate this and apply further optimization in the future.

result

ktock:
Merging this because all tasks completed.

PR #73, opened 2020-04-16
Re-resolve the connection when checking layers fail

ktock:
Currently `filesystem.Check` returns error when the checking layer fails,
without recovery. This error finally turns to Unavailable error on
snapshotter's side. There are several possible causes for the connection error
but some of them including creds update should be recoverable on the
filesystem's side by simply re-resolving the image ref/digest, without returning
Unavailable. This commit enables filesystem to try to recover the connection if
possible. This hopefully leads to higher availability for long-living layers.

This commit makes `Blob` agnostic about the detail of the connection and abstracts out it to a object `fetcher`. `fetcher` is created and registered to a `Blob`  by global `Resolver`. `Blob` refers to a `fetcher` object and uses it for doing remote-related works including fetching blobs. When `Blob.Check()` fails during `filesystem.Check()`, filesystem tries to refresh the `Blob` and tells `Resolver` to do so by calling `Resolver.Refresh(Blob)`. This re-resolves and re-create the `fetcher` for specified `Blob` and replace the `fetcher` of this `Blob` with newly resolved one.


script/integration/containerd/entrypoint.sh
```diff
@@ -70,25 +91,25 @@ cp /auth/certs/domain.crt /usr/local/share/ca-certificates
 update-ca-certificates
 retry docker login "${REGISTRY_HOST}:5000" -u "${DUMMYUSER}" -p "${DUMMYPASS}"
 
-echo "Preparing images..."
-gcrane cp ubuntu:18.04 "${REGISTRY_HOST}:5000/ubuntu:18.04"
-gcrane cp alpine:3.10.2 "${REGISTRY_HOST}:5000/alpine:3.10.2"
-PREFIX=/tmp/ctr/ make clean && \
-    PREFIX=/tmp/ctr/ make ctr-remote && \
-    install /tmp/ctr/ctr-remote /usr/local/bin
+echo "Insatalling snapshotter..."
```
AkihiroSuda, line 94: typo

script/integration/containerd/entrypoint.sh
```diff
@@ -70,25 +91,25 @@ cp /auth/certs/domain.crt /usr/local/share/ca-certificates
...
+mkdir -p /tmp/out
```
AkihiroSuda, line 95: nit: consider using `mktemp -d`. Not a huge deal for CI though.

script/integration/containerd/entrypoint.sh
```diff
@@ -101,36 +122,81 @@ if [ "${OK}" != "ok" ] ; then
...
+iptables -D OUTPUT -d "${REGISTRY_ALT_HOST_IP}" -j DROP
+
+echo "Disabling all registries and running container should fail"
+iptables -A OUTPUT -d "${REGISTRY_HOST_IP}","${REGISTRY_ALT_HOST_IP}" -j DROP
+iptables -L
+if ctr-remote run --rm --snapshotter=stargz "${REGISTRY_HOST}:5000/alpine:stargz" test tar -c /usr > /usr_dummy_fail.tar ; then
+    echo "All registries are disabled so this must be failed"
+    exit 1
+else
+    echo "Successfully failed running container"
```
AkihiroSuda, line 163: wut

AkihiroSuda: "Successfully failed" -> "Failed as expected"?

ktock: Sorry for the bad english:pray: Yes I meant "Failed as expected".

ktock:
Thanks for the review. Fixed all and ready for review on green.

PR #78, opened 2020-04-21
Ensure stargz snapshotter starts before containerd.service

ktock:


ktock:
Thanks!:smile:

tuananh:
Sorry to comment on this old PR @ktock. I thought snapshotter needs containerd in order to work?

why we start it after containerd.service?

PR #81, opened 2020-04-23
Use single range request for registries which don't support multi range request

ktock:
Some cloud-provided registries including GCR doesn't support multi range request (e.g. `range: bytes=0-3,5-9,16-20`)
For example, when we pull an image from GCR,
```console
# ctr-remote image rpull gcr.io/stargz-275102/ubuntu:18.04
fetching sha256:36d56a39... application/vnd.docker.distribution.manifest.v2+json
fetching sha256:b993a6c1... application/vnd.docker.container.image.v1+json
fetching sha256:ce77d9a4... application/vnd.docker.image.rootfs.diff.tar.gzip
fetching sha256:cc65679c... application/vnd.docker.image.rootfs.diff.tar.gzip
fetching sha256:dcc819e8... application/vnd.docker.image.rootfs.diff.tar.gzip
fetching sha256:1b677c82... application/vnd.docker.image.rootfs.diff.tar.gzip
```
We get 400 response because of multi range request.
```
DEBU[2020-04-23T02:27:15.607869311Z] failed to read layer                          digest="sha256:cc65679ceb37dd5d861fe76ced27bbca8897d3c73f9e5a4422491c5932837804" error="failed to parse stargz: error reading footer: unexpected status code on \"https://storage.googleapis.com/artifacts.stargz-275102.appspot.com/containers/images/sha256:cc65679ceb37dd5d861fe76ced27bbca8897d3c73f9e5a4422491c5932837804\": 400 Bad Request" mountpoint=/var/lib/containerd-stargz-grpc/snapshotter/snapshots/1/fs ref="gcr.io/stargz-275102/ubuntu:18.04"
```

This commit solves this issue by using a single request instead.
We might need tests with cloud-provided registries. I'll add them with another PR.

### DIFF
--- a/stargz/fs.go
+++ b/stargz/fs.go
@@ -184,6 +184,8 @@ func (fs *filesystem) Mount(ctx context.Context, mountpoint string, labels map[s
 	// will be stopped during the execution so this can avoid being disturbed for
 	// NW traffic by background tasks.
 	sr := io.NewSectionReader(readerAtFunc(func(p []byte, offset int64) (n int, err error) {
+		fs.backgroundTaskManager.DoPrioritizedTask()
+		defer fs.backgroundTaskManager.DonePrioritizedTask()
 		return blob.ReadAt(p, offset)
 	}), 0, blob.Size())
 	gr, root, err := reader.NewReader(sr, fs.fsCache)
--- a/stargz/remote/blob_test.go
+++ b/stargz/remote/blob_test.go
@@ -68,15 +68,49 @@ func TestReadAt(t *testing.T) {
 		"in_3_chunks_blob": sampleChunkSize * 3,
 		"in_max_size_blob": int64(len(sampleData1)),
 	}
-	cacheCond := map[string][]region{
-		"with_clean_cache": nil,
-		"with_edge_filled_cache": {
-			region{0, sampleChunkSize - 1},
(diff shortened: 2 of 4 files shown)


ktock:
We don't see significant performance degradation.
result

AkihiroSuda:
Can we try multi-range first and then fall back to single-range?

ktock:
> Can we try multi-range first and then fall back to single-range?

We can.
Committed the fallback version of the patch for benchmarking purpose.
We need to squash the commits before merge them.

ktock:
Ready for review.
result

AkihiroSuda:
Significant performance regression on glassfish:stargz?

ktock:
Ready for review.
Seems it was because we made each file accesses non-prioritized task in the past commit. Fixed it together.


result

AkihiroSuda:
Still slower compared to https://raw.githubusercontent.com/containerd/stargz-snapshotter/4d750b3ea358c64d77eb68ac4b9809459d3d7eea/docs/images/benchmarking-result-288c338.png ?

ktock:
Measured with that commit 288c338 on GA but glassfish:stargz is slower as well. https://github.com/containerd/stargz-snapshotter/pull/81/checks?check_run_id=614503013
Maybe it's because of the something outside of the codes.
We need to investigate the impact of internet condition and seek the way to mitigate the dispersion...

result

PR #83, opened 2020-04-25
Resolve layers in parallel

ktock:
Currently, resolving each layer in an image are done sequentially. But it can be done in parallel by passing all layer digests in the image to `Prepare` of this snapshotter.

This commit enables this by adding a new optional label `containerd.io/snapshot/remote/stargz.chain` for passing all digests in the target image and by resolving all digests in parallel.


ktock:
`Pull` performance becomes better.
result

PR #98, opened 2020-05-14
Fix potential data race

ktock:
Detected by the Go race detector (`-race` option).
This commit also enables the race detector at some points in our CI.


AkihiroSuda:
-race should not be enabled for benchmark

ktock:
It's enabled only in 'Optimize' and 'Integration' test.

PR #102, opened 2020-05-22
Define log message for failure of remote snapshot preparation

ktock:
Currently, we have two ways to check the success of remote snapshot
preparation. Integration test checks output messages from `ctr-remote`
command. CRI test checks the log from stargz daemon. We should unify them
because we want to keep the consideration the about log messages as small as
possible.

We use the log from stargz daemon for both tests. For preventing the message
from accidentally changed, this commit defines the message as `const` value and
adds some comments.

This commit also adds this check to the benchmarking script.

### DIFF
--- a/snapshot/snapshot.go
+++ b/snapshot/snapshot.go
@@ -37,6 +37,16 @@ import (
 const (
 	targetSnapshotLabel = "containerd.io/snapshot.ref"
 	remoteLabel         = "containerd.io/snapshot/remote"
+
+	// remoteSnapshotLogKey is a key for log line, which indicates whether
+	// `Prepare` method successfully prepared targeting remote snapshot or not, as
+	// defined in the following:
+	// - "true"  : indicates the snapshot has been successfully prepared as a
+	//             remote snapshot
+	// - "false" : indicates the snapshot failed to be prepared as a remote
+	//             snapshot
+	// - null    : undetermined
+	remoteSnapshotLogKey = "remote-snapshot-prepared"
 )
 
 // FileSystem is a backing filesystem abstraction.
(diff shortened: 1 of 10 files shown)


snapshot/snapshot.go
```diff
@@ -37,6 +37,11 @@ import (
 const (
 	targetSnapshotLabel = "containerd.io/snapshot.ref"
 	remoteLabel         = "containerd.io/snapshot/remote"
+
+	// logPrepareRemoteFailure is log message for indicating failure of
+	// preparing remote snapshot. This is used by tests in this project
+	// to make sure layers have been prepared as remote snapshots.
+	logPrepareRemoteFailure = "failed to prepare remote snapshot"
```
AkihiroSuda, line 44: consider using JSON lines rather than defining this kind of magic string

ktock:
I agree with your comment. Though the change in the new commit is a bit larger than the initial scope in this PR, it formats the entire log with JSON. This is because this kind of use-cases (i.e. querying information from logs) might occur in the future again and we want to avoid the same issue so I thought it's good timing for moving to JSON. LMK what you think.

snapshot/snapshot.go
```diff
@@ -37,6 +37,16 @@ import (
...
+
+	// remoteSnapshot is a key for log line, which indicates whether `Prepare`
+	// method successfully prepared targeting remote snapshot or not, as defined
+	// in the following:
+	// - "true"  : indicates the snapshot has been successfully prepared as a
+	//             remote snapshot
+	// - "false" : indicates the snapshot failed to be prepared as a remote
+	//             snapshot
+	// - null    : undetermined
+	remoteSnapshot = "containerd.io/snapshot/prepare/remote.snapshot"
```
AkihiroSuda, line 49: too long maybe

ktock: I thought to avoid name conflict because the key will be global but, yes I think so :thinking:

Issue #86, opened 2020-04-27
Migrate to go-fuse v2


AkihiroSuda:
https://github.com/containerd/stargz-snapshotter/blob/cc738d8ee6183945db474916be3e2d2baec1ad79/go.mod#L23

We should consider migrating to go-fuse v2: https://github.com/hanwen/go-fuse/releases

PR #103, opened 2020-05-26
Migrate to go-fuse V2

ktock:
Fixes: https://github.com/containerd/stargz-snapshotter/issues/86


cmd/ctr-remote/logger/logger.go
```diff
@@ -44,19 +45,23 @@ const (
...
+		AttrTimeout:     &timeSec,
+		EntryTimeout:    &timeSec,
+		NullPermissions: true,
+	})
+	if err := root.InitNodes(); err != nil {
+		return nil, errors.Wrap(err, "failed to init nodes")
+	}
+	server, err := fuse.NewServer(rawFS, mountPoint, &fuse.MountOptions{
+		AllowOther: true,             // allow users other than root&mounter to access fs
+		FsName:     "logger",         // name this filesystem as "logger"
```
AkihiroSuda, line 59: why?

ktock: About `FsName` field? It may be unnecessary. I'll fix it.

PR #108, opened 2020-06-12
 Use single Dockerfile for version management of containerd in tests

ktock:
Currently, each test has its own Dockerfile and manages the version of containerd and runc, etc. by its own, which is a bit costly for us in terms of version management.
This commit refactors it by defining a common Dockerfile which defines version of containerd and runc used in this project and which can be used as a base image for testing containers.  This commit also introduces "CinD"(Containerd In Docker) build stage (image) in that Dockerfile which can be used for easy demo or manual testing.


Dockerfile
```diff
@@ -12,30 +12,81 @@
...
+                                 apt-transport-https gnupg2 software-properties-common && \
+    curl -fsSL https://download.docker.com/linux/debian/gpg | apt-key add - && \
+    add-apt-repository \
+      "deb [arch=amd64] https://download.docker.com/linux/debian $(lsb_release -cs) stable" && \
+    apt-get update -y && apt-get --no-install-recommends install -y docker-ce-cli
+COPY --from=containerd-dev \
+     /out/bin/containerd \
+     /out/bin/containerd-shim \
+     /out/bin/containerd-shim-runc-v1 \
+     /out/bin/containerd-shim-runc-v2 \
```
AkihiroSuda, line 61: Probably you don't need old shims

script/benchmark/test.sh
```diff
@@ -18,29 +18,59 @@ set -euo pipefail
...
+    cd \${GOPATH}/src/github.com/google/go-containerregistry && \
+    git checkout 4b1985e5ea2104672636879e1694808f735fd214 && \
+    GO111MODULE=on go install github.com/google/go-containerregistry/cmd/crane
+
+COPY ./config/config.containerd.toml /etc/containerd/config.toml
+COPY ./config/config.stargz.toml /etc/containerd-stargz-grpc/config.toml
+
+ENV CONTAINERD_SNAPSHOTTER=""
+
+ENTRYPOINT [ "tail", "-f", "/dev/null" ]
```
AkihiroSuda, line 61: sleep infinity

Dockerfile
ktock: Thanks for the comment. Fixed it.

script/benchmark/test.sh
ktock: Thanks for the comment. Fixed it.

PR #109, opened 2020-06-16
Add `Unmount` API to the FileSystem

ktock:
Our `FileSystem` implementation holds a `map[string]*layer` for tracking the
state of each stargz mountpoint. Currently, `FileSystem` interface defined in
`snapshot` package doesn't have `Unmount` API and `syscall.Unmount` is directly
called to the mountpoint. Filesystem implementation asynchronously finalizes
that `map`, which is hard to be done safely and rather race condition will possibly
occur.

This is good timing for us to add `Unmount` method to the `FileSystem` API,
which delegates the actual `unmount` task to the filesystem implementation. This
also enables the filesystem to do necessary finalization before/after the
`unmount` syscall.

### DIFF
--- a/snapshot/snapshot_test.go
+++ b/snapshot/snapshot_test.go
@@ -415,6 +415,10 @@ func (fs *bindFs) Check(ctx context.Context, mountpoint string) error {
 	return nil
 }
 
+func (fs *bindFs) Unmount(ctx context.Context, mountpoint string) error {
+	return syscall.Unmount(mountpoint, 0)
+}
+
 func dummyFileSystem() FileSystem { return &dummyFs{} }
 
 type dummyFs struct{}
@@ -424,6 +428,10 @@ func (fs *dummyFs) Mount(ctx context.Context, mountpoint string, labels map[stri
 }
 
 func (fs *dummyFs) Check(ctx context.Context, mountpoint string) error {
+	return fmt.Errorf("dummy")
+}
+
+func (fs *dummyFs) Unmount(ctx context.Context, mountpoint string) error {
 	return fmt.Errorf("dummy")
 }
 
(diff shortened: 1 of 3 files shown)


snapshot/snapshot_test.go
```diff
@@ -415,6 +415,10 @@ func (fs *bindFs) Check(ctx context.Context, mountpoint string) error {
 	return nil
 }
 
+func (fs *bindFs) Unmount(ctx context.Context, mountpoint string) error {
+	return syscall.Unmount(mountpoint, 0)
```
AkihiroSuda, line 419: Do we want `MNT_FORCE` here? I'm not sure.

stargz/fs.go
```diff
@@ -425,6 +416,17 @@ func (fs *filesystem) check(ctx context.Context, l *layer) error {
...
 
+func (fs *filesystem) Unmount(ctx context.Context, mountpoint string) error {
+	fs.layerMu.Lock()
+	if _, ok := fs.layer[mountpoint]; !ok {
+		fs.layerMu.Unlock()
+		return fmt.Errorf("specified path %q isn't a mountpoint", mountpoint)
+	}
+	delete(fs.layer, mountpoint) // unregisters the corresponding layer
+	fs.layerMu.Unlock()
+	return syscall.Unmount(mountpoint, 0)
```
AkihiroSuda, line 427: ditto

ktock: It is possible that some of the server goroutines become not responding so let's use MNT_FORCE to abort the FUSE connection (https://www.kernel.org/doc/html/latest/filesystems/fuse.html#aborting-a-filesystem-connection). In the future, we might be able to consider to kill that specific hanging goroutine using channel, etc.
But let's don't care about test codes now.

Issue #114, opened 2020-07-01
[RFC] stargz image verification


ktock:
Since https://github.com/google/crfs/pull/37, TOCEntry contains the checksum (SHA256 digest) of regular files so we can use these digests for file verification once we verify TOCEntry itself in the chain of verification (manifest -> TOCEntry -> file digests). But currently, TOCEntry digest itself isn't contained in verifiable blobs (manifest, etc).

One of the straightforward ways is to store the digest of TOCEntry as annotations in manifest. This might require,

- [x] changes on `ctr-remote image optimize` command for containing the TOCEntry digest to the annotation
- [x] changes on the filesystem to verify file contents
- [x] ~changes on cri plugin for passing the annotations to stargz snapshotter during pull~ (annotations are automatically passed down to this snapshotter)
- [x] ~changes on google/crfs for making it return TOCEntry digest from the [`*stargz.Writer`](https://github.com/google/crfs/blob/71d77da419c90be7b05d12e59945ac7a8c94a543/stargz/stargz.go#L527-L546) and for making it verify that digest [during opening the stargz file](https://github.com/google/crfs/blob/71d77da419c90be7b05d12e59945ac7a8c94a543/stargz/stargz.go#L65-L106).~(added [wrapper](https://github.com/containerd/stargz-snapshotter/blob/e96a16531aa2de2ad9c6c1f02ec01abe0da53de0/stargz/verify/verify.go))
- [x] update images on dockerhub for making them verifiable
- [x] add docs

The fourth one can also be done by manually parsing raw stargz blob outside of the stargz lib (in `optimize` command and in the filesystem) but eventually it should be better to upstream it.

ktock:
We've done fundamental verification features. Let's close this thread and do follow-ups in other ones when needed.

PR #116, opened 2020-07-06
Extend stargz for content verification

ktock:
Related: https://github.com/containerd/stargz-snapshotter/issues/114

Current stargz image has two problems towards content validation:
- The digest of TOC JSON isn't stored in verifiable image manifests
- Content digests are stored in file granularity (not in chunk granularity)

By solving them, we can construct a chain of verification (i.e. manifest -> TOC JSON -> chunk).

This commit solves them in backwards-compatible ways as the following.

For the first one, this commit adds an annotation `containerd.io/snapshot/stargz/toc.digest` to the image manifest for storing the digest of TOC JSON.

For the second one, this commit extends the TOCEntry with an additional (and optional) field `ChunkDigest` for storing chunk-level content digest. Stargz lib chunks a large file to smaller "chunk" entries (4MiB by default). In stargz snapshotter, we leverage this and fetch a large file in chunk granularity, selectively. However, stargz lib __doesn't__ calculate the digest for each chunk. Instead, it calculates the digest for the entire file and stores this to`TOCEntry.Digest` of the first chunk and `TOCEntry.Digest` of other chunks are kept empty (see also: https://github.com/google/crfs/pull/37). For keeping backwards-compatibility with stargz, this commit adds a new field for containing chunk-level digests instead of changing the semantics of `TOCEntry.Digest`.

This commit is for the archive format and the converter command. I'll post another PR for the verification functionality on the filesystem side leveraging this format.(or, should it be contained in this PR?)

### DIFF
--- a/stargz/verify/verify.go
+++ b/stargz/verify/verify.go
@@ -1,0 +1,232 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package verify
+
+import (
+	"archive/tar"
+	"bytes"
+	"compress/gzip"
+	"crypto/sha256"
+	"encoding/json"
+	"fmt"
+	"io"
+	"strconv"
+
+	"github.com/google/crfs/stargz"
+	"github.com/pkg/errors"
+)
+
+const (
+	// TOCJSONDigestAnnotation is an annotation for image manifest. This stores the
+	// digest of the TOC JSON
+	TOCJSONDigestAnnotation = "containerd.io/snapshot/stargz/toc.digest"
+)
+
+// jtoc is a TOC JSON schema which contains additional fields for chunk-level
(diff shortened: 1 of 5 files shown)


stargz/verify/verify.go
```diff
@@ -0,0 +1,225 @@
...
+	// digest of the TOC JSON
+	TOCJSONDigestAnnotation = "containerd.io/snapshot/stargz/toc.digest"
+)
+
+// jtoc is a TOC JSON schema which contains additional fields for chunk-level
+// content verification. This is still backward compatible to the official
+// definition:
+// https://github.com/google/crfs/blob/71d77da419c90be7b05d12e59945ac7a8c94a543/stargz/stargz.go
+type jtoc struct {
+	Version int         `json:"version"`
```
AkihiroSuda, line 44: Needs documentation

stargz/verify/verify.go
```diff
@@ -0,0 +1,225 @@
...
+	"github.com/pkg/errors"
+)
+
+const (
+	// TOCJSONDigestAnnotation is an annotation for image manifest. This stores the
+	// digest of the TOC JSON
+	TOCJSONDigestAnnotation = "containerd.io/snapshot/stargz/toc.digest"
+)
+
+// jtoc is a TOC JSON schema which contains additional fields for chunk-level
```
AkihiroSuda, line 39: Make sure to document our extension in markdown as well

stargz/verify/verify.go
AkihiroSuda: Seems unused field?

ktock: Added documentation. `Version` is unused in this package but we need it for keeping the backwards-compatibility with stargz.

stargz/verify/verify.go
ktock: Can we add it after the patches on filesystem side (= after making whole things e2e)?

PR #117, opened 2020-07-11
Enable filesystem to verify stargz blob contents

ktock:
Related to: #114, #116

Currently, stargz filesystem doesn't verify the layer contents. This PR enables the filesystem to verify:

- TOC JSON of the stargz layer to mount, and
- each content contained in that layer.

The following briefly describes commits contained in this PR.

### TOC JSON verification functionality in `verify` package (c40f0477dcc78f365415e7e13b8627597d1c18d1)

This commit extends `verify` package with a verification functionality. A new function `verify.StargzTOC` checks if the given stargz blob has a valid TOC JSON. "Valid" TOC JSON here means:

- digest of the TOC JSON matches the one passed through the argument, and
- each TOCEntry of regfile/chunk in the TOC contains the digest of the content.

Once the TOC JSON is verified, this function returns a set of `"github.com/opencontainers/go-digest".Verifier` keyed by `stargz.TOCEntry` contained in the TOC JSON.

### Layer verification on the filesystem (cd9d084f15db4851db7cc1983cacce86da3c82fe)

Since #116, an eStargz image contains expected digests of TOC JSON in the manifest as annotations. Once the manifest is verified (on containerd side), the expected digest of TOC JSON written in the manifest is passed down to this snapshotter for each mount operation. In this commit, using this digest and `verify.StargzTOC` mentioned in the above, filesystem verifies the TOC JSON of the stargz blob to mount and gets a set of `go-digest.Verifier` keyed by `stargz.TOCEntry`. For each fetching regular file or chunk from registry, `reader.Reader` verifies the downloaded content using `Verifier` corresponding to the `TOCEntry`. So we can connect the chain of verification from the image manifest to each file contents lazily fetched from the registry.

For legacy stargz images that don't contain the TOC JSON digest annotations, filesystem allows to mount them if either of the following conditions meets,

- `allow_no_verification` is set to `true` in the config.toml of this snapshotter
- label `"containerd.io/snapshot/remote/stargz.skipverify"` is appended to that layer

This commit adds a flag to `ctr-remote image rpull` command to allow the filesystem to deal with legacy stargz images, using the second method in the above.

### Remaining TODOs:
- Enabling verification in benchmarking tests
  - We also need to update images on DockerHub for making them verifiable so it should be a separated PR.
- More discussion on how to deal with the verification failure
  - Currently, `EIO` will simply be returned to the filesystem user on verification failure.
  - Should we prohibit all further filesystem operation for this layer as well?
  - Should we immediately unmount the layer? But maybe state dir should still be visible for debugging?
  - I personally think this can be a separated PR because it'll require heavy refactoring on the codes around filesystem and cache. But let me know if this should be done also in this PR.


go.mod
```diff
@@ -22,7 +22,7 @@ require (
 	github.com/google/go-containerregistry v0.0.0-20200425101607-48f605c3b60a
 	github.com/hanwen/go-fuse/v2 v2.0.3
 	github.com/imdario/mergo v0.3.9 // indirect
-	github.com/opencontainers/go-digest v1.0.0-rc1.0.20180430190053-c9281466c8b2 // indirect
+	github.com/opencontainers/go-digest v1.0.0-rc1.0.20180430190053-c9281466c8b2
```
AkihiroSuda, line 25: not directly related to this PR, but you should bump up this to 1.0.0

stargz/verify/verify.go
```diff
@@ -120,70 +122,157 @@ func NewVerifiableStagz(sgz *io.SectionReader) (newSgz io.Reader, jtocDigest str
...
+		return nil, fmt.Errorf("verifier for offset=%d,size=%d hasn't been registered",
+			ce.Offset, ce.ChunkSize)
+	}
+	return d.Verifier(), nil
+}
+
+// tocInfo is a set of information related to footer and TOC JSON in a stargz
+type tocInfo struct {
+	jtoc   *jtoc         // decoded TOC JSON
+	digest string        // SHA256 digest of TOC JSON. formed as "sha256:abcd1234..."
```
AkihiroSuda, line 211: consider using this type: https://pkg.go.dev/github.com/opencontainers/go-digest?tab=doc#Digest

stargz/verify/verify.go
```diff
@@ -120,70 +122,157 @@ func NewVerifiableStagz(sgz *io.SectionReader) (newSgz io.Reader, jtocDigest str
...
+	offset int64         // offset of TOC JSON in the stargz blob
+	footer io.ReadSeeker // stargz footer
+}
+
 // extractTOCJSON extracts TOC JSON from the given stargz reader, with avoiding
 // scanning the entire blob leveraging its footer. The parsed TOC JSON contains
 // additional fields usable for chunk-level content verification.
-func extractTOCJSON(sgz *io.SectionReader) (toc *jtoc, tocOffset int64, footer io.ReadSeeker, err error) {
+func extractTOCJSON(sgz *io.SectionReader) (toc tocInfo, err error) {
```
AkihiroSuda, line 219: `tocInfo` -> `*tocInfo` so that we can return nil

stargz/verify/verify.go
```diff
@@ -120,70 +122,157 @@ func NewVerifiableStagz(sgz *io.SectionReader) (newSgz io.Reader, jtocDigest str
...
+		return tocInfo{}, errors.Wrap(err, "failed to get TOC JSON tar entry")
 	} else if h.Name != stargz.TOCTarName {
-		return nil, 0, nil, fmt.Errorf("invalid TOC JSON tar entry name %q; must be %q",
+		return tocInfo{}, fmt.Errorf("invalid TOC JSON tar entry name %q; must be %q",
 			h.Name, stargz.TOCTarName)
 	}
-	toc = new(jtoc)
-	if err := json.NewDecoder(tr).Decode(&toc); err != nil {
-		return nil, 0, nil, errors.Wrap(err, "failed to decode TOC JSON")
+	dgst := sha256.New()
```
AkihiroSuda, line 256: Use OCI `Digster`

go.mod
ktock: Thanks for the comment. Bumped up in this PR.

stargz/verify/verify.go
ktock: Fixed patches to use `digest.Digest` instead of `string`.

stargz/verify/verify.go
ktock: Fixed it.

stargz/verify/verify.go
ktock: Fixed patches to use `Digester` instead of directly using `sha256`, etc.

PR #118, opened 2020-07-12
Fix variable type in config at overview.md

koxu1996:
Type of option 'insecure' is bool, not string.


ktock:
@koxu1996 Please sign your commit from the next PR.

koxu1996:
@ktock Ok, I will remember that :)

PR #125, opened 2020-08-03
Add doc about content verification

ktock:
Related: #114


docs/verification.md
```diff
@@ -0,0 +1,58 @@
+# Content Verification in eStargz
+
+The goal of the verifiability in eStargz is to make all files' metadata and contents in an eStargz layer verifiable.
```
AkihiroSuda, line 3: Definition of "verify"?

docs/verification.md
```diff
@@ -0,0 +1,58 @@
+# Content Verification in eStargz
```
AkihiroSuda, line 1: link this file from `stargz-estargz.md`

docs/verification.md
```diff
@@ -0,0 +1,58 @@
...
+`ctr-remote images optimize` command in this project creates the verifiable eStargz image by default.
+
+## Content verification in Stargz Snapshotter
+
+Stargz Snapshotter verifies eStargz layers leveraging the above extensions.
+However, as mentioned in the above, the verification of other image component including the manifests is out-of-scope of the snapshotter.
+So when this snapshotter mounts an eStargz layer, the manifest that references this layer must be verified in the containerd in advance and the TOC's digest written in the manifest (as an annotation `containerd.io/snapshot/stargz/toc.digest`) must be passed down to this snapshotter.
+
+If the manifest doesn't contain `containerd.io/snapshot/stargz/toc.digest` annotation, verification can't be done for that layer so stargz snapshotter reports an error and doesn't mount it.
+You can bypath this check only if both of the following conditions meet.
```
AkihiroSuda, line 49: bypass

docs/verification.md
ktock: Through the series of commits related #114, what I aimed to was enabling:

"ensuring the downloaded metadata and contents of all files are the expected ones, based on the calculated OCI digests"
And I called it "verification" so far. 

So the first line should be:
```
The goal of the content verification in eStargz is to ensure the downloaded metadata and contents of all files are the expected ones, based on the calculated OCI digests.
```

Please tell me if I'm missing something or there is something to be added/fixed in the definition and/or the implementation.

ktock: And thanks for the comment! :pray:

PR #130, opened 2020-08-18
Eliminate URL information in error logs

ktock:
This commit eliminates URL information from error logs for avoiding potential security risk.
In the future, we may want additional debug logs containing URL information for development purpose but let's keep it separated patch.

### DIFF
--- a/stargz/remote/resolver.go
+++ b/stargz/remote/resolver.go
@@ -202,8 +202,7 @@ func (r *Resolver) resolve(ref, digest string) (*fetcher, int64, error) {
 		// Get size information
 		size, err = getSize(url, tr)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


stargz/remote/resolver.go
```diff
@@ -176,7 +176,7 @@ func (r *Resolver) resolve(ref, digest string) (*fetcher, int64, error) {
 	for _, h := range hosts {
 		// Parse reference
 		if h.Host == "" || strings.Contains(h.Host, "/") {
-			rErr = errors.Wrapf(rErr, "host %q: mirror must be a domain name", h.Host)
```
AkihiroSuda, line 179: hostname might be safe to leak?

Issue #145, opened 2020-09-12
Allow to disable fetching layers in background


koxu1996:
I have rather specific use case of _StarGZ_: I use it to save disk space, because I am running 64 nodes on my local machine. I don't really care about runtime performance, so fetching layers in background is undesirable behavior for me. Currently I use forked version, but I though maybe such option could be useful for someone else.

I was thinking about adding new config option and wrapping [background fetching](https://github.com/containerd/stargz-snapshotter/blob/master/stargz/fs.go#L324-L356) with something like: `if !fs.nolazypull { ... }`

I could prepare pull request, but would like to introduce such change?

ktock:
SGTM. Can you open a PR?

koxu1996:
Thanks :smiley:

PR #150, opened 2020-09-20
Allow to disable pulling layers in background

koxu1996:
Simple enhancement described in https://github.com/containerd/stargz-snapshotter/issues/145: lazy pulling can be disabled by setting `nolazypull = true` in _config.toml_.

I tested this option by executing `cat /.stargz-snapshotter/*` inside container - when it is enabled, then _fetchedSize_ is not growing up.

### DIFF
--- a/stargz/config/config.go
+++ b/stargz/config/config.go
@@ -29,6 +29,7 @@ type Config struct {
 	PrefetchSize        int64  `toml:"prefetch_size"`
 	PrefetchTimeoutSec  int64  `toml:"prefetch_timeout_sec"`
 	NoPrefetch          bool   `toml:"noprefetch"`
+	NoBackgroundFetch   bool   `toml:"no_background_fetch"`
 	Debug               bool   `toml:"debug"`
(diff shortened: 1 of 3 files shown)


stargz/config/config.go
```diff
@@ -29,6 +29,7 @@ type Config struct {
 	PrefetchSize        int64  `toml:"prefetch_size"`
 	PrefetchTimeoutSec  int64  `toml:"prefetch_timeout_sec"`
 	NoPrefetch          bool   `toml:"noprefetch"`
+	NoLazyPull          bool   `toml:"nolazypull"`
```
ktock, line 32: Because each file content is still lazily pulled on-demand, `nolazypull ` isn't proper term here. `NoBackgroundFetch`  or something should be better

ktock (commented):
Thanks! Sign your commit, please.

koxu1996:
@ktock Fixed

ktock (approved):

Issue #129, opened 2020-08-18
optimize image with --stargz-only get an error "digest of TOC JSON must be passed"


cy-zheng:
cmd:
```
ctr-remote image optimize --plain-http --stargz-only golang:1.14.7 {private-registry}/chenyu-test/golang:1.14.7
ctr-remote images rpull --user='{UNAME:PWD}' --plain-http {private-registry}/chenyu-test/golang:1.14.7
```

log output:
```
Aug 17 22:15:45 ava-staging-k8s-stress-test-worker09 containerd-stargz-grpc[2346]: {"error":"digest of TOC JSON must be passed","key":"default/178/extract-707423280-RKQt sha256:943d4a5efe67b44aab74815a110a231c5018e3dac04a4a4f1b9cf826d8000a8e","level":"debug","msg":"failed to prepare remote snapshot","parent":"","remote-snapshot-prepared":"false","time":"2020-08-17T22:15:45.584216442-07:00"}
```

cmd:
```
root@ava-staging-k8s-stress-test-worker09:~# ctr-remote run --rm -t --snapshotter=stargz {private-registry}/chenyu-test/golang:1.14.7 test /bin/bash
root@ava-staging-k8s-stress-test-worker09:/go# cat /.stargz-snapshotter/*
cat: '/.stargz-snapshotter/*': No such file or directory
```

ktock:
Stargz (non-eStargz) image doesn't support [content verification](https://github.com/containerd/stargz-snapshotter/blob/master/docs/verification.md) so the snapshotter doesn't mount it by default.
You need the following configuration for mounting stargz (non-eStargz) layers (see also [the doc](https://github.com/containerd/stargz-snapshotter/blob/master/docs/verification.md#content-verification-in-stargz-snapshotter)).

- set `allow_no_verification = true` in config.toml of stargz snapshotter, and
- specify `--skip-content-verify` to `ctr-remote images rpull` command.

AkihiroSuda:
Is this issue closable? Do we need more clear docs?

koxu1996:
@ktock Is there possibility to use non-eStargz image with KinD? I added `allow_no_verification = true` but I have no clue where is place for `--skip-content-verify` extra argument.

ktock:
@koxu1996 Possible solutions:

- Manually adding `containerd.io/snapshot/remote/stargz.skipverify` to your non-eStargz image.
[...]

PR #152, opened 2020-09-24
Allow to disable verification completely

koxu1996:
Added flag for disabling verification completely - mentioned in https://github.com/containerd/stargz-snapshotter/issues/129#issuecomment-698046132

### DIFF
--- a/docs/verification.md
+++ b/docs/verification.md
@@ -51,6 +51,8 @@ You can bypass this check only if both of the following conditions meet.
 - `allow_no_verification = true` is specified in `config.toml` of stargz snapshotter, and
 - the content descriptor of this layer has an annotation `containerd.io/snapshot/remote/stargz.skipverify` (the value will be ignored).
 
+The other way is to disable verification completely by setting `disable_verification = true` in `config.toml` of stargz snapshotter.
+
 On mounting a layer, stargz snapshotter fetches this layer's TOC from the registry.
 Then it verifies the TOC by recaluculating the digest and comparing it with the one passed from containerd (written in the manifest).
 If the TOC is successfully verified, then the snapshotter mounts this layer using the metadata stored in the TOC.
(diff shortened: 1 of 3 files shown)


docs/verification.md
```diff
@@ -51,6 +51,8 @@ You can bypass this check only if both of the following conditions meet.
 - `allow_no_verification = true` is specified in `config.toml` of stargz snapshotter, and
 - the content descriptor of this layer has an annotation `containerd.io/snapshot/remote/stargz.skipverify` (the value will be ignored).
 
+The other way is to disable verification completely by setting `disable_verification = true` in `config.toml` of stargz snapshotter. This option is necessary when working with stargz (non-eStargz) format.
```
ktock, line 54: > The other way is to disable verification completely by setting `disable_verification = true` in `config.toml` of stargz snapshotter.

sounds good to me.

> This option is necessary when working with stargz (non-eStargz) format.

This isn't accurate because we can currently work with stargz (non-eStargz) format when we use `ctr-remote` command, even without `disable_verification`. Maybe we don't need this sentence here.

stargz/fs.go
```diff
@@ -279,6 +281,10 @@ func (fs *filesystem) Mount(ctx context.Context, mountpoint string, labels map[s
 		// necessary for layer verification.
 		l.skipVerify()
 		log.G(ctx).Warningf("No verification is held for layer")
+	} else if fs.disableVerification {
+		// Verification is disabled completely
+		l.skipVerify()
+		log.G(ctx).Debugf("Verification forcefully skipped")
```
ktock, line 287: `disable_verification` disables verification completely, so this check should come to the top of the `if` chain and should disable the verification even for eStargz images.

ktock (commented):
Thanks!

docs/verification.md
koxu1996: I removed the sentence.

stargz/fs.go
koxu1996: Ok, moved to the top.

ktock (approved):

Issue #156, opened 2020-10-01
Too many goroutines causing full CPU for large images


strigazi:
Hello,

(So if you click by accident ENTER on the title you have an issue created.)

When doing ctr-remote i rpull for `docker.io/stargz/python:3.7-esgz` around 25K goroutines get created, then they go away (after ~30s).

For larger images like `docker.io/strigazi/ciadm:v0.3.9-esgz-bash-version` (OR `ghcr.io/strigazi/ciadm:v0.3.9-esgz-bash-version`) I get close to 1M goroutines that after 10mins they were still there. All 16 cores of my host get consumed from the snapshotter at 100%. I have an even larger image, I will try push it to docker.io, I have it locally for now.

I can pass you the profile from pprof if you need.

strigazi:
cc @rochaporto @ktock

strigazi:
From pprof I think that the routines get created  here [0] and they close only when all files are traversed.

[0] https://github.com/containerd/stargz-snapshotter/blob/8f2e7ec8468e67ed7251709af7841c6ca5ac64d6/stargz/reader/reader.go#L150

ktock:
@strigazi Thanks for reporting this. Stargz snapshotter aggressively pulls and decompresses a layer in background aside from fetching contents on demand. These goroutines do this "background pull/decompression" per file. 
I wrote a patch to avoid too many goroutines there.

If you don't want snapshotter to pull/decompress layers in background, you can disable it by adding `no_background_fetch = true` to snapshotter's `config.toml`. File contents won't be fetched until access to that file occurs.

strigazi:
> @strigazi Thanks for reporting this. Stargz snapshotter aggressively pulls and decompresses a layer in background aside from fetching contents on demand. These goroutines do this "background pull/decompression" per file.
> I wrote a patch #157 to avoid too many goroutines there.
> 
> If you don't want snapshotter to pull/decompress layers in background, you can disable it by adding `no_background_fetch = true` to snapshotter's `config.toml`. File contents won't be fetched until access to that file occurs.

I tried this config option, but I see the same behaviour (all cores consumed, too many goroutines)

strigazi:
Fixed by #157

PR #157, opened 2020-10-02
Avoid launching too many goroutines during caching layer

ktock:
#156

Currently `reader.Cache()` creates goroutines for each file contained in the
layer. This can lead to too many goroutines and possibly consume too much
resource.

This commit fixes it by limiting the number of goroutines using semaphore.

cc: @strigazi

### DIFF
--- a/stargz/reader/reader.go
+++ b/stargz/reader/reader.go
@@ -29,6 +29,7 @@ import (
 	"fmt"
 	"io"
 	"path/filepath"
+	"runtime"
 	"sync"
 
 	"github.com/containerd/stargz-snapshotter/cache"
@@ -36,6 +37,7 @@ import (
 	"github.com/google/crfs/stargz"
 	"github.com/pkg/errors"
 	"golang.org/x/sync/errgroup"
+	"golang.org/x/sync/semaphore"
 )
 
 type Reader interface {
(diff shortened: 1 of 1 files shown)


stargz/reader/reader.go
```diff
@@ -133,22 +134,28 @@ func (gr *reader) Cache(opts ...CacheOption) (err error) {
 		filter = cacheOpts.filter
 	}
 
-	return gr.cacheWithReader(root, r, filter, cacheOpts.cacheOpts...)
+	eg, egCtx := errgroup.WithContext(context.Background())
+	eg.Go(func() error {
+		return gr.cacheWithReader(egCtx, eg, semaphore.NewWeighted(4),
```
AkihiroSuda, line 139: > semaphore.NewWeighted(4)

Should this corresponds to GOMAXPROCS?

AkihiroSuda: Curious, is it really allowed to recursively call `eg.Go` from `eg.Go`?

strigazi:
Thanks for the promt reply and super fast proposal!

The situation is better with this patch.
good:
* goroutines stay low stay stable after a rpull of `ghcr.io/strigazi/ciadm:v0.3.9-esgz-bash-version` (less than 200, most of them for fuse functions).
* I see CPU at 100% on 1 of the 16 cores (before it was 100% on all cores i.e. >1500% on top).

bad :(
* After some time, the snapshotter craches with
```
runtime: goroutine stack exceeds 1000000000-byte limit
runtime: sp=0xc027f10330 stack=[0xc027f10000, 0xc047f10000]
fatal error: stack overflow
```

ktock:
@strigazi Thanks for trying this patch!

> After some time, the snapshotter craches

What's the stack trace showing?

This might be related to the recursive `eg.Go` call. (But this seems allowed pattern according to https://github.com/golang/go/issues/27837#issuecomment-612163144)
Another possibility is that the recursive call of gr.cacheWithReader is too deep because of the deep directory hierarchy in that layer.

strigazi:
> @strigazi Thanks for trying this patch!
> 
> > After some time, the snapshotter craches
> 
> What's the stack trace showing?
> 
> This might be related to the recursive `eg.Go` call. (But this seems allowed pattern according to [golang/go#27837 (comment)](https://github.com/golang/go/issues/27837#issuecomment-612163144))
> Another possibility is that the recursive call of gr.cacheWithReader is too deep because of the deep directory hierarchy in that layer.

https://gist.github.com/strigazi/56c0857f77c0a6153bd3602e31dd6646

stargz/reader/reader.go
ktock: Thanks for reviewing this. 

> Should this correspond to GOMAXPROCS?

Fixed it on the latest push.

> Curious, is it really allowed to recursively call eg.Go from eg.Go?

According to https://github.com/golang/go/issues/27837#issuecomment-612163144, this seems an allowed pattern.

ktock:
@strigazi Thanks for providing the log. That panic was because of an infinite recursive call of `gr.cacheWithReader` caused by `./` entry on the root which points to the root directory itself. I updated the patch to avoid this issue.

```
$ cat ./img/f560fdf781d7ff7e85bf15288fcb9941ed3a7f7434e024f62dd488b1deb452ee/layer.tar | tar -xO stargz.index.json | jq '.entries[].name' | head -n 10
".prefetch.landmark"
"./"
"./root/"
"./root/.tcshrc"
"./root/anaconda-ks.cfg"
"./root/original-ks.cfg"
"./root/.bash_profile"
"./root/.cshrc"
"./root/.bash_logout"
"./root/.bashrc"
```

strigazi:
Thanks @ktock , it looks good now!

Low number of goroutines for the images that I tried, CPU utilization is normal.

AkihiroSuda (approved):

PR #158, opened 2020-10-04
Add scope configration

ktock:
Authorizer in containerd lib requires scope configuration via context. 
Currently our resolver misses this, which leads to fail to connect to registries (e.g. ghcr.io).
This commit fixes the issue.

### DIFF
--- a/stargz/remote/resolver.go
+++ b/stargz/remote/resolver.go
@@ -32,6 +32,7 @@ import (
 	"mime"
 	"mime/multipart"
 	"net/http"
+	"net/url"
 	"path"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


stargz/remote/resolver.go
```diff
@@ -128,6 +133,9 @@ func newFetcher(hosts []docker.RegistryHost, refspec reference.Spec, digest stri
 			tr = &transport{
 				inner: tr,
 				auth:  h.Authorizer,
+				// Specify pull scope
+				// TODO: The scope generator function in containerd (github.com/containerd/containerd/remotes/docker/scope.go) should be exported and used here.
```
AkihiroSuda, line 137: Open a PR

AkihiroSuda (approved):

Issue #149, opened 2020-09-17
remove dependency on github.com/google/go-containerregistry 


AkihiroSuda:
https://github.com/moby/buildkit/pull/1666#discussion_r489907766

We could just use the containerd libs for registry stuff

tonistiigi:
Another question would be how to get mirrors/tls configuration etc. I'm not sure how authentication works atm, possibly needs some updates as well.

ktock:
Now snapshotter plugin has containerd's RegistryHost-based registry logic.
https://github.com/containerd/stargz-snapshotter/pull/153

PR #161, opened 2020-10-13
Allow client to configure blob URLs through RegistryHost configuration

ktock:
Related to: #149
This commit adds a new field `RefAlias` to the registry configuration. This
enables the client to configure the destination URL of layers more flexibly. For
example, the client can change the URL of a layer blob on refresh, based on the
connectivity.

This implies that a snapshot (or layer) shouldn't be tied to a single image
reference and we might need more refactoring on the registry logic but let's
keep this separated from this patch.


ktock:
Closing in favor of #163.

PR #163, opened 2020-10-17
Move label-related logic into a separated package

ktock:
Previous: containerd/stargz-snapshotter#161
Related: https://github.com/moby/buildkit/pull/1733

This commit separates labels-related logic from the filesystem core. This introduces a new package `source`(`github.com/containerd/stargz-snapshotter/stargz/source`) and defines a logic to convert snapshot labels into blob sources information `source.Source`. 

```golang
type Source struct {
	// Hosts is a registry configuration where this blob is stored.
	Hosts docker.RegistryHosts

	// Name is an image reference which contains this blob.
	Name reference.Spec

	// Target is a descriptor of this blob.
	Target ocispec.Descriptor

	// Manifest is an image manifest which contains the blob. This will
	// be used by the filesystem to pre-resolve some layers contained in
	// the manifest.
	// Currently, only layer digests (Manifest.Layers.Digest) will be used.
	Manifest ocispec.Manifest
}
```

`source` package defines an interface of that converter (`GetSources`) , and contains the default implementation based on `containerd.io/snapshot/remote/stargz...` labels (which is previously embedded in the filesystem core).

```golang
type GetSources func(labels map[string]string) (source []Source, err error)
```

Now the filesystem uses `source.GetSources` for acquiring blob source information. By default, it uses the implementation by `source` package but it accepts any user-defined ones through a constructor option `WithGetSources`.

`containerd-stargz-grpc` uses the following `source.GetSources` implementations.
- a parser of default labels `containerd.io/snapshot/remote/stargz...` (implemented in `source` package).
- a parser of labels `containerd.io/snapshot/cri...` passed from CRI, which previously embedded in the filesystem core.

Another note comes with this change is that the filesystem is now agnostic about (source-related) labels. `source.GetSources` implementations aren't required to use these labels so they can provide blob source information leveraging other media instead of these labels (e.g. using manifest-like JSON data + content store).


cmd/containerd-stargz-grpc/main.go
```diff
@@ -192,3 +198,17 @@ func hostsFromConfig(cfg ResolverConfig, keychain authn.Keychain) remote.Registr
...
+
+func sources(ps ...source.GetSources) source.GetSources {
+	return func(labels map[string]string) (source []source.Source, _ error) {
+		allErr := fmt.Errorf("cannot parse labels")
+		for _, p := range ps {
+			src, err := p(labels)
+			if err == nil {
+				return src, nil
+			}
+			allErr = errors.Wrapf(allErr, "%v;", err)
```
AkihiroSuda, line 210: Can we use https://github.com/hashicorp/go-multierror ?

ktock: Thanks. Fixed to use go-multierrror.

AkihiroSuda (approved):

Issue #184, opened 2020-12-03
Change docs and benchmarks to more focus on eStargz


ktock:
eStargz is a format which is based on stargz by Google CRFS project and comes with optimization and verification. Based on this context, currently, our docs describes both of stargz and eStargz.
 However, recently we are focusing more on eStargz and maintaining docs and benchmarks for both of stargz and eStargz is a bit costly. For making the project easier to maintain, let's fix them to focus more on eStargz (of course, with preserving necessary credits to stargz by CRFS project).

TODOs:
- [x] #186 fix benchmark to measure the following
  - legacy image
  - eStargz without optimization (not stargz)
  - eStargz with optimization
- [x] #189 fix docs to more focus on eStargz
- [x] #190 fix the implementation to make it compliant to the docs

cc: @AkihiroSuda

PR #186, opened 2020-12-03
Fix benchmark to use non-optimized eStargz instead of stargz

ktock:
Related: #184

This fixes benchmarking scripts for using eStargz (without optization) instead
of stargz. The non-optimized images are built by `--no-optimize` option of
`ctr-remote image optimize`.

From: https://github.com/ktock/stargz-snapshotter/actions/runs/398452630

result


AkihiroSuda (approved):
Promising result 👍

PR #192, opened 2020-12-09
Add an option to ignore context canceled.

mc256:
So I want to run the container a bit longer until it finishes all the
work for optimization. This option allows the container to wait until a
keyboard signal (`Ctrl-C`) or a signal from other programs.

Added this option to `ctr-remote i optimize` command:
```
 --wait-on-signal             ignore context cancel and keep the container running until it receives signal (Ctrl + C) sent manually
```

Signed-off-by: Jun Lin Chen webmaster@mc256.com

### DIFF
--- a/cmd/ctr-remote/sampler/sampler.go
+++ b/cmd/ctr-remote/sampler/sampler.go
@@ -70,7 +70,7 @@ func Run(ctx context.Context, bundle string, config v1.Image, opts ...Option) er
 	}
 
 	// run the container
-	if err := runContainer(ctx, bundle); err != nil {
+	if err := runContainer(ctx, bundle, opt.waitOnSignal); err != nil {
 		return errors.Wrap(err, "failed to run containers")
 	}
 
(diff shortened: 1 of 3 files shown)


cmd/ctr-remote/sampler/sampler.go
```diff
@@ -361,21 +361,31 @@ func runContainer(ctx context.Context, bundle string) error {
...
 	defer signal.Stop(sc)
-	select {
-	case <-done:
-		return nil
-	case <-ctx.Done():
-		log.G(ctx).Info("context canceled")
-	case <-sc:
-		log.G(ctx).Info("signal detected")
+	if ignoreCtxCancel {
+		log.G(ctx).Debugf("press Ctrl+C to terminate the container %q", id)
```
ktock, line 372: Can we make it `Infof`? This information seems important for user.

mc256: Sure!

ktock (approved):
Thanks!

PR #198, opened 2020-12-16
Split optimizer

AkihiroSuda:
Decouple non-CLI codes from CLI.

No substantial code change.


AkihiroSuda:
We should also split `Optimize()` to several steps, but that will be another PR.

ktock (approved):

Issue #200, opened 2020-12-16
Expose an estargz.Option for compression level


mattmoor:
https://pkg.go.dev/github.com/containerd/stargz-snapshotter@v0.2.0/estargz#Option

I'm guessing that as a compatible gzip format, the notion of compression level should persist.  The standard library exposes this level, and the following enums: https://godoc.org/compress/gzip#pkg-constants

In GGCR we generally bias towards `gzip.BestSpeed`, but we expose an option for this in our own utilities around gzip.  It would be nice to preserve this optionality with estargz.

ktock:
Though the [footer](https://github.com/containerd/stargz-snapshotter/blob/6ed3e72ebbc0d5f67f2b3eb16787f7e1a2890f05/docs/stargz-estargz.md#footer) must be gzip.NoCompression, we can expose compression level of other components (entries and toc).

ktock:
I meant that I totally agree with the suggestion :smile:

ktock:
Closing this as we support `WithCompression` option since  #201.

PR #201, opened 2020-12-16
Enable to pass compression level to gzip lib

ktock:
#200
https://github.com/kubernetes-sigs/kind/issues/1981

This enables estargz lib client to specify compression level of gzip lib by adding a new option `WithCompressionLevel`.
According to eStargz spec, footer must be 51 bytes so footer's compression level must not be configurable (i.e. gzip.NoCompression).


AkihiroSuda:
> According to eStargz spec, footer must be 51 bytes so footer's compression level must not be configurable (i.e. gzip.NoCompression).

This should be explicitly noted in the spec

AkihiroSuda (approved):

Issue #203, opened 2020-12-16
Move the prioritized files to an estargz.Option


mattmoor:
Thinking about API in downstream go-containerregistry, I find myself wanting to be able to expose `estargz.Option` to folks directly, but one things I'd like them to be able to specify (that's currently separate) is the prioritized file list.

ktock:
Now prioritized files is moved to an option since #206.

PR #206, opened 2020-12-17
Move prioritized files argument to option

ktock:
#203

This moves  prioritized files argument to option.
There are cases where input tar entiy names are prefixed with "./" (which is
still valid) but we shouldn't expose this difference in the option.
So this commit also adds a simple normalization (trimming "./" prefix).

### DIFF
--- a/estargz/build.go
+++ b/estargz/build.go
@@ -45,6 +45,7 @@ import (
 type options struct {
 	chunkSize        int
 	compressionLevel int
+	prioritizedFiles []string
 }
 
 type Option func(o *options)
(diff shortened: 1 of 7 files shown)


estargz/build.go
```diff
@@ -65,12 +66,21 @@ func WithCompressionLevel(level int) Option {
 	}
 }
 
+// WithPrioritizedFiles option specifies the list of prioritized files.
+// These files must be a complete path relative to "/" (e.g. "foo/bar",
+// "./foo/bar")
+func WithPrioritizedFiles(files []string) Option {
+	return func(o *options) {
```
AkihiroSuda, line 73: This should verify `files` and return `error`.

Can be a follow-up PR.

AkihiroSuda (approved):
LGTM but a single nit

estargz/build.go
ktock: I'll work on it in the following up PR.

PR #212, opened 2020-12-19
Propagate not-found error during sort

ktock:
https://github.com/google/go-containerregistry/pull/879#issuecomment-748398008

Currently, even if a path specified in the prioritized files list isn't found in
the tar archive, no error is propagated. But this possibly result in the eStargz
file unexpected by the client so we should propagate this error to the client.

cc @AkihiroSuda @mattmoor

### DIFF
--- a/estargz/build.go
+++ b/estargz/build.go
@@ -296,7 +296,9 @@ func sortEntries(in io.ReaderAt, prioritized []string) ([]*entry, error) {
 	// Sort the tar file respecting to the prioritized files list.
 	sorted := &tarFile{}
 	for _, l := range prioritized {
-		moveRec(l, intar, sorted)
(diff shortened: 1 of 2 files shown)


AkihiroSuda (approved):

mattmoor:
Does this require all of the prioritized files to appear in the layer?  This will probably break the way I'm playing with this in `crane optimize` because the same list is passed to all layers 🤔

ktock:
@mattmoor How about introducing a new option like `WithPrioritizeBestEffort(missedFiles *[]string)` that makes `estargz.Build` ignore not-found error but record all missed files to the passed `missedFiles` slice?

mattmoor:
Yeah, I'd probably makes `missedFiles ...string` tho 😉

Issue #97, opened 2020-05-13
Support default path to the configuration file


ktock:
Stargz snapshotter daemon (`containerd-stargz-grpc`) currently doesn't have the default path to the configuration file. So users need to explicitly specify the path using `--config` option.
We should have the default path for the configuration file ([containerd also has the default](https://github.com/containerd/containerd/blob/f13ba8f2f2fd27fc2172453434938170e7807149/docs/man/containerd-config.toml.5.md#synopsis)). `/etc/containerd-stargz-grpc/config.toml` seems good for the default path.

ohkinozomu:
@ktock 
Hi. I've created PR about this. Could you review it?

PR #214, opened 2020-12-19
Add defaultConfigPath

ohkinozomu:
resolve https://github.com/containerd/stargz-snapshotter/issues/97

In addition to `script/integration/containerd/entrypoint.sh`, I found `--config=/etc/containerd-stargz-grpc/config.toml` hardcoded in several other files.
However, I thought it would be better to make them explicit in some cases, so I have not removed all of them.

Signed-off-by: Nozomu Ohki <nozomunoise@gmail.com>

### DIFF
--- a/cmd/containerd-stargz-grpc/main.go
+++ b/cmd/containerd-stargz-grpc/main.go
@@ -42,14 +42,15 @@ import (
 )
 
 const (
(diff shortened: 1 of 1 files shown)


ktock (approved):
Thanks!

ktock (changes requested):
Please squash commits.

ktock (approved):
Thanks!

PR #216, opened 2020-12-20
Allow some of prioritized files are missed during sort

ktock:
https://github.com/containerd/stargz-snapshotter/pull/212#issuecomment-748483624

This commit adds a new option WithAllowPrioritizeNotFound. This makes Build
continue the execution even if some of prioritized files specified by
WithPrioritizedFiles option aren't found in the input tar. Instead, this records
all missed file names to the passed slice.

### DIFF
--- a/estargz/build.go
+++ b/estargz/build.go
@@ -44,17 +44,19 @@ import (
 )
 
 type options struct {
-	chunkSize        int
-	compressionLevel int
-	prioritizedFiles []string
+	chunkSize              int
+	compressionLevel       int
+	prioritizedFiles       []string
+	missedPrioritizedFiles *[]string
 }
 
-type Option func(o *options)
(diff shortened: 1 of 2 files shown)


ktock:
@mattmoor  WDYT?

mattmoor:
I see, so this would collect the missed files, so that the caller knows what files were missed?

ktock:
Yes. So caller can optimize the layer in best-effort and know which files are missed.  This information seems needed for indicating some warnings on the caller's side according to https://github.com/google/go-containerregistry/pull/879#issuecomment-748398008 ?

mattmoor:
Yeah I think that'll be perfect for aggregating across layers and images and surfacing back to the user what was missing.

AkihiroSuda (approved):

Issue #217, opened 2020-12-20
snapshotter doesn't like paths in ko images


mattmoor:
In github.com/google/ko built images, the paths we use within the tarball seem to be causing problems for the estargz snapshotter.  These images run fine against a standard containerd, so it would be good to harden this snapshotter to support them:

```
Dec 20 21:36:45 ubuntu containerd-stargz-grpc[1179]: {"error":"failed to cache prefetched layer: invalid child path \"/ko-app\"; must be child of \"\"","level":"debug","mountpoint":"/var/lib/containerd-stargz-grpc/snapshotter/snapshots/28/fs","msg":"failed to prefetched layer","time":"2020-12-20T21:36:45.919228015Z"}
...
Dec 20 21:36:52 ubuntu containerd-stargz-grpc[1179]: {"error":"invalid child path \"/var\"; must be child of \"\"","level":"debug","mountpoint":"/var/lib/containerd-stargz-grpc/snapshotter/snapshots/27/fs","msg":"failed to fetch whole layer","time":"2020-12-20T21:36:52.422678402Z"}
```

You can produce an image with `ko` + estargz by:
1. clone github.com/google/ko
2. `go install ./cmd/ko`
3. `GGCR_EXPERIMENT_ESTARGZ=1 KO_DOCKER_REPO=docker.io/{username} ko publish -B ./cmd.foo`
4. `kubectl run foo --image={digest from above}`

mattmoor:
It looks like the distroless images (run through `crane optimize`) exhibit a similar error:
```
Dec 20 22:04:02 ubuntu containerd-stargz-grpc[1179]: {"error":"invalid child path \"/var\"; must be child of \"\"","level":"debug","mountpoint":"/var/lib/containerd-stargz-grpc/snapshotter/snapshots/46/fs","msg":"failed to fetch whole layer","time":"2020-12-20T22:04:02.521706097Z"}
```

mattmoor:
If I manually force `ko` to use relative paths for files within the tarball then it seems to stop complaining about that layer, but this is likely going to be a problem for a whole class of images, so it'd be best to try and fix this.

mattmoor:
cc @itsmurugappan @ktock @AkihiroSuda

ktock:
@mattmoor Thanks for reporting this.
There seems several issues for lazy pulling distroless images. I'll fix them today.

[...]

PR #218, opened 2020-12-21
estargz: Clean entry names when read and unmarshal blob

ktock:
https://github.com/containerd/stargz-snapshotter/issues/217#issuecomment-748710444

Currently, when estargz lib reads an eStargz, it normalizes the entry by
trimming the prefix "./".

But this isn't enough because there are a lot of exceptions including
"/foo/bar", "../foo/bar", etc. From the library client's perspective
(e.g. snapshotter), it's complex to take care of all these differences when
interacting with unmarshaled eStargz blob with `estargz.Reader`.

For example, when the client gets a TOCEntry from `estargz.Reader.Lookup(path
string)`, it needs to prefix the path to be lexically equal to the corresponding
tar entry (so it needs to prefix that path with "/" if the corresponding tar
entry is prefixed by "/"). Additionally, when the client gets a TOCEntry
(e.g. by `estargz.TOCEntry.ForeachChild()`), the names are possibly prefixed
with "/", "./", "../", "", etc and it needs to handle all of these cases
properly.

This commit solves this by fixing estargz lib to handle entry paths with
cleaning them internally when it provides an unmarshaled view of an eStargz blob
through `estargz.Reader`. They are cleaned as relative to root, and estargz lib
APIs (e.g. `estargz.Reader.Lookup()`) also clean paths passed by the
client. This enables the client to read each TOCEntry.Name as a clean path and
enables to query TOCEntry without thinking about the differences of prefixes in
the marshalled TOC.

Note that paths are cleaned only in the `estargz.Reader` but not in
`estargz.Writer`. So TOC JSON created by estargz.Writer still contains
TOCEntry.Name identical to tar entry name (possibly prefixed with "/", "../",
"./" etc).

This commit also fixed tests to run for each prefix patterns in "", "./", "/"
and "../".

### DIFF
--- a/estargz/build.go
+++ b/estargz/build.go
@@ -32,7 +32,6 @@ import (
 	"io/ioutil"
 	"os"
 	"path"
-	"path/filepath"
 	"runtime"
 	"strings"
 	"sync"
@@ -393,7 +392,7 @@ func importTar(in io.ReaderAt) (*tarFile, error) {
 				return nil, errors.Wrap(err, "failed to parse tar file")
 			}
 		}
-		switch strings.TrimPrefix(cleanRootRelative(h.Name), "/") {
+		switch cleanEntryName(h.Name) {
 		case PrefetchLandmark, NoPrefetchLandmark:
 			// Ignore existing landmark
 			continue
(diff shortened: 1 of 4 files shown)


AkihiroSuda (approved):

AkihiroSuda:
needs rebase

PR #225, opened 2020-12-22
Use path vs. filepath, so that the library cleans things properly on Windows

mattmoor:


mattmoor:
I can't restart the failed action, but it looks like a transient network issue.

mattmoor:
The downstream windows build passes with this pulled in: https://travis-ci.org/github/google/go-containerregistry/jobs/751069345

ktock:
Could you add a comment on the changed line something like "this prevents the separator to be replaced unintentionally"?

mattmoor:
I added a comment roughly describing things, but essentially since we're using `/` then `path.Clean` is the right thing to use across all platforms.

`filepath.Clean` is the right thing to use if we were using `filepath.Separator`, but I don't think that's right here because the platform we're running on isn't necessarily indicative of the separator we should use.

I have very little experience with Windows containers, but IIRC that container runtime supports `/` -> `\` so that we can consistently use `/` everywhere.  That may be inaccurate, but IDK how much y'all have tested this with windows?

ktock:
Thanks for adding the comment!

> I added a comment roughly describing things, but essentially since we're using / then path.Clean is the right thing to use across all platforms.
> 
> filepath.Clean is the right thing to use if we were using filepath.Separator, but I don't think that's right here because the platform we're running on isn't necessarily indicative of the separator we should use.

I totally agree with you.
Currently, we test only on linux but let's expand the test cases to other platforms once the library gets stable.

ktock (approved):

Issue #226, opened 2020-12-22
context deadline exceeded while prefetching


mattmoor:
Not sure why, but there is a deadline here:
https://github.com/containerd/stargz-snapshotter/blob/af03a142bbeefb0e07e915e4e8fdedc082e5039b/fs/remote/blob.go#L270

I see the following errors in the snapshotter's logs:
```
Dec 22 18:09:38 ubuntu containerd-stargz-grpc[290174]: {"error":"failed to prefetch layer: failed to get chunks: context deadline exceeded","level":"debug","mountpoint":"/var/lib/containerd-stargz-grpc/snapshotter/snapshots/535/fs","msg":"failed to prefetched layer","time":"2020-12-22T18:09:38.248309104Z"}
```

This is on a fairly resource constrained system/connection (RPi4), and some of the go binaries being prefetched are somewhat large (tens of MB).

Things seem to keep going, but I'd guess they are less efficient than they should be 😞

PR #228, opened 2020-12-23
Make timeout for fetching chunks configurable

ktock:
Resolves: #226

Currently, the timeout for fethcing chunks is 30s but this is too short on some
environments.
This commit make this configurable on `config.toml` (with `fetching_timeout`
option) and extended the default timeout (300s).

cc: @mattmoor

### DIFF
--- a/fs/config/config.go
+++ b/fs/config/config.go
@@ -53,9 +53,10 @@ type Config struct {
 }
 
 type BlobConfig struct {
-	ValidInterval int64 `toml:"valid_interval"`
-	CheckAlways   bool  `toml:"check_always"`
-	ChunkSize     int64 `toml:"chunk_size"`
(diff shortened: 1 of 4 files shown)


mattmoor (approved):
thanks!

fs/config/config.go
```diff
@@ -56,6 +56,7 @@ type BlobConfig struct {
 	ValidInterval int64 `toml:"valid_interval"`
 	CheckAlways   bool  `toml:"check_always"`
 	ChunkSize     int64 `toml:"chunk_size"`
+	FetchTimeout  int64 `toml:"fetching_timeout"`
```
AkihiroSuda, line 59: FetchTimeoutSeconds.
Or can we use time.Duration?

AkihiroSuda (approved):

PR #237, opened 2021-01-05
Add MaximumConcurrency to config.toml (for stargz-grpc)

mc256:
Increasing the number of concurrent background tasks (connections) may speed up the loading speed if the network condition is good. I propose to add a parameter to adjust the number of concurrent background tasks.

Signed-off-by: Jun Lin Chen <webmaster@mc256.com>

### DIFF
--- a/fs/config/config.go
+++ b/fs/config/config.go
@@ -44,6 +44,7 @@ type Config struct {
 	Debug               bool   `toml:"debug"`
 	AllowNoVerification bool   `toml:"allow_no_verification"`
 	DisableVerification bool   `toml:"disable_verification"`
+	MaxConcurrency      int64  `toml:"max_concurrency"`
 
 	// BlobConfig is config for layer blob management.
 	BlobConfig `toml:"blob"`
--- a/fs/fs.go
+++ b/fs/fs.go
@@ -83,6 +83,7 @@ const (
 	stateDirName              = ".stargz-snapshotter"
 	defaultResolveResultEntry = 100
 	defaultPrefetchTimeoutSec = 10
+	defaultMaxConcurrency     = 2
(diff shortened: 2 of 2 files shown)


fs/fs.go
```diff
@@ -161,7 +161,7 @@ func NewFilesystem(root string, cfg config.Config, opts ...Option) (_ snbase.Fil
 		layer:                 make(map[string]*layer),
 		resolveResult:         lru.New(resolveResultEntry),
 		blobResult:            lru.New(resolveResultEntry),
-		backgroundTaskManager: task.NewBackgroundTaskManager(2, 5*time.Second),
+		backgroundTaskManager: task.NewBackgroundTaskManager(cfg.MaximumConcurrency, 5*time.Second),
```
ktock, line 164: We should have the default value = 2.

ktock (changes requested):
Thanks! ~~What is the use-case of this patch?~~ Nevermind. I've just read your PR doc now.

script/demo/config.stargz.toml
```diff
@@ -1,4 +1,5 @@
 disable_verification = true
+maximum_concurrency = 2
```
ktock, line 2: We don't need this as the default = 2.

fs/config/config.go
```diff
@@ -44,6 +44,7 @@ type Config struct {
 	Debug               bool   `toml:"debug"`
 	AllowNoVerification bool   `toml:"allow_no_verification"`
 	DisableVerification bool   `toml:"disable_verification"`
+	MaximumConcurrency  int64  `toml:"maximum_concurrency"`
```
ktock, line 47: Should be something like `MaxBackgroundConcurrency`.

ktock (approved):

Issue #230, opened 2020-12-24
FUSEless file access monitor (inotify, fanotify)


AkihiroSuda:
[The current FUSE-based monitor implementation](https://github.com/containerd/stargz-snapshotter/blob/7e30a23808d1c17c7ff05e1d693bb8e3aa66ec7d/converter/optimizer/logger/logger.go) has several issues:
- Does not work without CAP_SYS_ADMIN, i.e., does not work inside Kaniko/Makisu
- Hard to support lazy-pulling during monitoring
- ...

So I suggest adding inotify implementation.

The drawback of inotify is that it can't be used for a rootfs with a lot of directories.
So we can consider adding fanotify implementation as well, though it requires privileges.

AkihiroSuda:
cc @ktock @mattmoor

ktock:
SGTM :+1:

Issue #223, opened 2020-12-22
optimizer: imageio should use containerd native API


AkihiroSuda:
https://github.com/containerd/stargz-snapshotter/blob/f045b42c062fac08ea440afc080de49765e2252d/converter/optimizer/imageio/imageio.go#L38-L44

@ktock 
For better integration with containerd, can we switch away from ggcr API into containerd content/image store API?
Especially, the converted image should be written to a containerd local image store without pushing.

AkihiroSuda:
draft PR: https://github.com/containerd/stargz-snapshotter/pull/224

AkihiroSuda:
Reopening. `ctr-remote images convert` is implemented, but `ctr-remote images optimize` still uses ggcr implementation.

ktock:
After #238 is merged, I'll work on the refactoring of `ctr-remote image optimize` based on native converter + analyzer + containerd's registry lib. I'm thinking about how we can keep this working without running containerd daemon.

AkihiroSuda:
>  I'm thinking about how we can keep this working without running containerd daemon.

Why not run the daemon?

Anyway, we could implement a static optimizer that parses ELF and shell scripts to estimate file dependencies without running any container at all, though its robustness would be limited.

jonjohnsonjr:
> Anyway, we could implement a static optimizer that parses ELF and shell scripts to estimate file dependencies without running any container at all, though its robustness would be limited.

This sounds really interesting, especially if it's possible (and easy) for you to keep something like this decoupled from containerd.

PR #238, opened 2021-01-06
Add --optimize option to ctr-remote image convert

ktock:
#230
#223

This commit adds `--optimize` option to ctr-remote image convert for enabling to
optimize the input image. This adds new package `analyzer` for analyzing layers
and dertermining prioritized files. Currently, FUSE-based analyzer is only
implemented but we can have other implementations (e.g. FUSEless analyzer) in
the future.

The following changes will come after this PR:
- having more analyzer e.g. FUSEless analyzer, static analyzer, containerd-based analyzer, etc.
- having converter command runnable without containerd daemon
  - modifying `ctr-remote image optimize` based on native converter + containerd's registry lib?


cmd/ctr-remote/commands/convert.go
```diff
@@ -66,6 +75,14 @@ When '--all-platforms' is given all images in a manifest list must be available.
 			Usage: "eStargz chunk size",
 			Value: 0,
 		},
+		cli.BoolFlag{
+			Name:  "reuse",
```
AkihiroSuda, line 79: estargz flags should have `estargz-` prefix as in existing flags.

cmd/ctr-remote/commands/convert.go
```diff
@@ -85,19 +102,31 @@ When '--all-platforms' is given all images in a manifest list must be available.
 			Name:  "all-platforms",
 			Usage: "exports content from all platforms",
 		},
-	},
-	Action: func(context *cli.Context) error {
+	}, samplerFlags...),
```
AkihiroSuda, line 105: I don't like this design.
`ctr images convert` should remain pure static image converter.
`ctr optimize` can have exec flags as in the current revision, but I think it should be reimplemented with `ctr image mount + ctr run + ctr images convert`.

ktock: Sure. I'll reconsider the design of `ctr optimize` to make it containerd-based.

ktock:
Closing in favor of #240.

PR #243, opened 2021-01-19
Bump go to 1.15

### LINKED ISSUE #202: update Go to 1.15
In #197 we tried to bump up Go from 1.13 to 1.15, but we failed because of the following FUSE error during `make benchmark`:

> writer: Write/Writev failed, err: 2=no such file or directory. opcode: INTERRUPT

Needs analysis and fix.

ktock:
Fixes #202 

When the snapshotter is compiled with especially the recent version of go, we
sometimes (or frequently) see that go-fuse prints `Write/Writev failed` log.
Here, I quote the log with few lines of debug logs around it.

```
2021/01/19 14:09:43 rx 76: FLUSH n2 {Fh 1}
2021/01/19 14:09:43 tx 76:     OK
...
2021/01/19 14:09:43 rx 77: INTERRUPT n0 {ix 76}
2021/01/19 14:09:43 tx 77:     11=resource temporarily unavailable
2021/01/19 14:09:43 writer: Write/Writev failed, err: 2=no such file or directory. opcode: INTERRUPT
```

This warning comes when the filesystem tries to write the response to INTERRUPT
(EAGAIN, in this case) to the kernel [1].

When a filesystem operation of the client from userspace (`76: FLUSH`, in this
case) is interrupted by a signal, INTERRUPT request (`77: INTERRUPT` targetting
`76: FLUSH`, in this case) is sent to the FUSE server. In this case, there is a
race condition according to the kernel doc:

(Quoted from [2])

```
It is also possible that there's a race between processing the
original request and its INTERRUPT request.  There are two possibilities:

  1) The INTERRUPT request is processed before the original request is
     processed

  2) The INTERRUPT request is processed after the original request has
     been answered

If the filesystem cannot find the original request, it should wait for
some timeout and/or a number of new requests to arrive, after which it
should reply to the INTERRUPT request with an EAGAIN error.  In case
1) the INTERRUPT request will be requeued.  In case 2) the INTERRUPT
reply will be ignored.
```

As far as we can observe, we are currently seeing `Write/Writev failed` error on
case 2 where we writes EAGAIN to the kernel. In the above log excerpt,
`77: INTERRUPT` is processed after `76: FLUSH` is returned with `OK`. Though
there is no document about errno in this case, FUSE seems to return ENOENT if
the interrupted request is not found on its side[3].

As a conclusion, this is an expected behaviour according to the kernel doc so we
can consider this phenomenon as harmless.

P.S. Similar phenomenon is observed on the internet[4], and they also conclude
this phenomenon as harmless.

[1] https://github.com/hanwen/go-fuse/blob/09a3c381714cf1011fb2d08885f29896cd496a0c/fuse/server_linux.go#L15
[2] https://www.kernel.org/doc/Documentation/filesystems/fuse.txt
[3] https://github.com/torvalds/linux/blob/219d54332a09e8d8741c1e1982f5eae56099de85/fs/fuse/dev.c#L1858-L1866
[4] https://bugzilla.redhat.com/show_bug.cgi?id=1763208#c17

### DIFF
--- a/cmd/containerd-stargz-grpc/main.go
+++ b/cmd/containerd-stargz-grpc/main.go
@@ -19,6 +19,7 @@ package main
 import (
 	"context"
 	"flag"
+	golog "log"
 	"net"
 	"net/http"
 	"os"
@@ -68,6 +69,10 @@ func main() {
 		ctx    = log.WithLogger(context.Background(), log.L)
 		config Config
 	)
+	// Streams log of standard lib (go-fuse uses this) into debug log
+	// Snapshotter should use "github.com/containerd/containerd/log" otherwize
+	// logs are always printed as "debug" mode.
+	golog.SetOutput(log.G(ctx).WriterLevel(logrus.DebugLevel))
 
 	// Get configuration from specified file
 	if _, err := toml.DecodeFile(*configPath, &config); err != nil && !(os.IsNotExist(err) && *configPath == defaultConfigPath) {
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,7 +19,7 @@ ARG CNI_PLUGINS_VERSION=v0.9.0
 # Legacy builder that doesn't support TARGETARCH should set this explicitly using --build-arg.
 # If TARGETARCH isn't supported by the builder, the default value is "amd64".
 
-FROM golang:1.13-buster AS golang-base
+FROM golang:1.15-buster AS golang-base
 
 # Build containerd
 FROM golang-base AS containerd-dev
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/containerd/stargz-snapshotter
 
-go 1.13
+go 1.15
 
 require (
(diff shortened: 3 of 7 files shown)


ktock:
Changed the test to ignore `writer: Write/Writev failed, err: 2=no such file or directory. opcode: INTERRUPT` error. Let me check if this change works on the GithubActions.

AkihiroSuda:
Can we just demote the log level to `debug` so that they won't appear in the CI log?

cmd/containerd-stargz-grpc/main.go
```diff
@@ -68,6 +69,10 @@ func main() {
 		ctx    = log.WithLogger(context.Background(), log.L)
 		config Config
 	)
+	// Streams log of standard lib (go-fuse uses this) into debug log
+	// Snapshotter should use "github.com/containerd/containerd/log" otherwize
```
AkihiroSuda, line 73: typo is negligible, so merging

AkihiroSuda (approved):

PR #256, opened 2021-02-04
Make sure the snapshotter is up-and-running during test

ktock:
Recently `Integration` test is flaky and sometimes fails with the following error:

```
failed to prepare extraction snapshot "extract-888733096-M1HL sha256:645c926ab234a37fce3b0d3f9de28718b25fdfe19537e6f58c01e25561116ccf": connection error: desc = "transport: Error while dialing dial unix /run/containerd-stargz-grpc/containerd-stargz-grpc.sock: connect: connection refused": unavailable
```

This seems because tests sometimes access to stargz snapshotter before it becomes up-and-running.
This commit fixes this issue.

### DIFF
--- a/script/integration/containerd/entrypoint.sh
+++ b/script/integration/containerd/entrypoint.sh
@@ -111,7 +111,10 @@ function reboot_containerd {
     fi
     retry ls "${REMOTE_SNAPSHOTTER_SOCKET}"
     containerd --log-level debug --config=/etc/containerd/config.toml &
-    retry ctr version
+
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):
I approve this, but this looks weird. I think the right way in the long term is to expose the plugin status in the introspection API. Or use systemd notification.

ktock:
> I think the right way in the long term is to expose the plugin status in the introspection API. Or use systemd notification.

Thanks for the suggestion. I'll work on it.

Issue #253, opened 2021-02-01
Improve tests and release


ktock:
Related: https://github.com/kubernetes-sigs/kind/issues/1981

Currently, we release only amd64 binary that tested against the latest containerd release. Aiming to facilitate the integration around stargz snapshotter, we can consider the following improvements.

- [x] #260 **release non-amd64 binraies as well**
  - ~~candidates: amd64, arm64, ppc64le~~
  - candidates: amd64, arm, arm64, ppc64le, s390x
    -  the architecture set Kubernetes ships (https://kubernetes.io/docs/setup/release/notes/)

- [x] #257 **run tests against containerd's nightly build as well**
  - this makes sure stargz snapshtoter works with the latest but non-released containerd.

- [x] #255 **slim down the size of binary & dependencies**
  - For enabling lazy pulling on containerd, [`containerd-stargz-grpc` binary and fuse package need to be installed at least](https://github.com/containerd/stargz-snapshotter/blob/0d28df31109e88f522417e46a5ca77eb381dfe13/Dockerfile#L95-L102) (aside from containerd).
    - `containerd-stargz-grpc` (built with Go 1.15) is about 44MB
    - `fuse3` pkg is about 1MB (so it can be considered harmless) https://packages.ubuntu.com/focal/fuse3
  - we should use `-w`, `-s` linker option (`ldflags`) for `containerd-stargz-grpc`; it slims down to 32MB
  - (if needed) reducing Go module dependencies will be needed but it might not be easy.
  - (if needed) should we consider using UPX ?

cc @AkihiroSuda

BenTheElder:
If possible: I would suggest the architecture set Kubernetes ships might be a better list. While KIND has a few things only targeting these architectures currently, most things target the full set.

ktock:
Thanks for the suggestion. Updated the issue description.

ktock:
Working on CI improvements (#257) that helps us catching up changes on containerd closely and quickly making fixes when some of the recent changes on containerd cause incompatibility.

After some consideration, I think the integration tests of stargz snapshotter + containerd should also run in KinD CI everytime it bumps containerd to un-released version. I've thought about guaranteeing that stargz snapshotter works arbitrary versions (commits) of containerd but this isn't easy because there are so many possible combinations of versions of containerd & snapshotter.

PR #260, opened 2021-02-09
Release cross builds

ktock:
#253

Release cross builds

This commit adds the following cross builds

- amd64
- arm
- arm64
- ppc64le
- s390x

Note that the test is currently done only against amd64 binaries.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,7 +10,10 @@ env:
 jobs:
   build:
     runs-on: ubuntu-20.04
-    name: Release
+    name: Build
+    strategy:
+      matrix:
+        arch: ["amd64", "arm-v7", "arm64", "ppc64le", "s390x"]
     env:
       OUTPUT_DIR: ${{ github.workspace }}/out
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -21,10 +24,28 @@ jobs:
       run: |
         mkdir ${OUTPUT_DIR}
         RELEASE_TAG="${GITHUB_REF##*/}"
-        TAR_FILE_NAME="stargz-snapshotter-${RELEASE_TAG}-linux-amd64.tar.gz"
+        TAR_FILE_NAME="stargz-snapshotter-${RELEASE_TAG}-linux-${{ matrix.arch }}.tar.gz"
```
AkihiroSuda, line 27: BuildKit and nerdctl uses `-linux-arm-v7.tgz` convention for ARM

Dockerfile
```diff
@@ -41,6 +41,7 @@ RUN apt-get update -y && apt-get install -y libseccomp-dev && \
 
 # Build stargz snapshotter
 FROM golang-base AS snapshotter-dev
+ARG GOARCH
```
AkihiroSuda, line 44: use TARGETARCH

.github/workflows/release.yml
ktock: Fixed to use `arm-v7` identifier.

Dockerfile
ktock: Fixed this.

AkihiroSuda (approved):
We could simplify the flow using builds --platform (that's why I suggested using TARGETARCH), but it can be worked out later, so LGTM

Issue #259, opened 2021-02-08
Make modules import-friendly


ktock:
Currently, stargz snaphsotter repo contains two go modules:

- `github.com/containerd/stargz-snapshotter`
- `github.com/containerd/stargz-snapshotter/estargz`

In our go.mod, `github.com/containerd/stargz-snapshotter` refers `github.com/containerd/stargz-snapshotter/estargz` using the following `replace` directive:

```
require github.com/containerd/stargz-snapshotter/estargz v0.0.0-00010101000000-000000000000
...
replace github.com/containerd/stargz-snapshotter/estargz => ./estargz
```

Here, `v0.0.0-00010101000000-000000000000` is not a valid version and this requires projects importing `github.com/containred/stargz-snapshotter` to specify the `replace` directive pointing to the valid one.

Example on BuildKit:

```
replace github.com/containerd/stargz-snapshotter/estargz => github.com/containerd/stargz-snapshotter/estargz v0.0.0-20201217071531-2b97b583765b
```

Furthermore, projects that indirectly depend on stargz snapshotter on their go.mod also need to specify this kind of directive, *even if they doesn't actually use codes from this repo* (e.g. https://github.com/tonistiigi/binfmt/pull/20).

We should eliminate the necessity to specify replace directive.

Additionally, we also should tag `github.com/containerd/stargz-snapshotter/estargz` not only `github.com/containerd/stargz-snapshotter` (the latest is `v0.3.0`).

Go module support to tag the submodule using the tag formatted as `<module-name>/<version>`.

https://golang.org/ref/mod#vcs-version

> If a module is defined in a subdirectory within the repository, that is, the module subdirectory portion of the module path is not empty, then each tag name must be prefixed with the module subdirectory, followed by a slash. For example, the module golang.org/x/tools/gopls is defined in the gopls subdirectory of the repository with root path golang.org/x/tools. The version v0.4.0 of that module must have the tag named gopls/v0.4.0 in that repository.

The version of `estargz` lib should be same as `github.com/containerd/stargz-snapshotter` pkg.

cc @tonistiigi @AkihiroSuda

AkihiroSuda:
SGTM

PR #261, opened 2021-02-09
[RFC] Bump estargz lib to a vaild oldest version

ktock:
#259

This commit eliminates the need of specifying `replace` directive in the
consumer project's go.mod.

Currently, our tests and releases are done against stargz snapshotter that
imports the local (i.e. the same commit of) `estargz` library.

```
replace github.com/containerd/stargz-snapshotter/estargz => ./estargz
```

Here, we don't want to force consumer projects (and their all downstream
projects) to specify `replace` directive but `require` directive doesn't support
to specify the local directory.

For solving this issue, this commit let the consumers manually specify the
version of `github.com/containerd/stargz-snapshotter/estargz` to the same commit
as `github.com/containerd/stargz-snapshotter` in their go.mod as an [indirect
requirement](https://golang.org/ref/mod#go-mod-file-require).

This commit intensionally bumps `estargz` lib to the "oldest" version to make
sure that the build fails if the version of
`github.com/containerd/stargz-snapshotter/estargz` is not specified explicitly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,6 +156,15 @@ root@8dab301bd68d:/# ls
 bin  boot  dev  etc  home  lib  lib64  media  mnt  opt  proc  root  run  sbin  srv  sys  tmp  usr  var
 ```
 
+## Importing Stargz Snapshotter as go module
+
+Currently, Stargz Snapshotter repository contains two Go modules as the following and both of them need to be imported.
+
+- `github.com/containerd/stargz-snapshotter`
+- `github.com/containerd/stargz-snapshotter/estargz`
+
+Please make sure you import the both of them and they point to *the same commit version*.
+
 ## Project details
 
 Stargz Snapshotter is a containerd **non-core** sub-project, licensed under the [Apache 2.0 license](./LICENSE).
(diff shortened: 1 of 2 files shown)


AkihiroSuda:
Why oldest

ktock:
Initially, I thought that can encourage the pkg consumer to override the version of `estargz` to the same commit as `github.com/containerd/stargz-snapshotter` (because build always fails when it's not explicitly overridden) but now I don't. 

Fixed to import the latest version and added a doc to emphasize that both of `github.com/containerd/stargz-snapshotter` and `github.com/containerd/stargz-snapshotter/estargz` should be imported with making sure that they have the same version.

AkihiroSuda:
I guess we had to bump up this to v0.4.0 before tagging v0.4.0 ?

ktock:
Though consumers can manually override `estargz` to v0.4.0 by their own, this is not ideal as you said...

```
require (
	github.com/containerd/stargz-snapshotter v0.4.0
	github.com/containerd/stargz-snapshotter/estargz v0.4.0 // indirect
)
```
Let's bump this to `v0.4.1` and tag `v0.4.1`.

Issue #272, opened 2021-02-19
CI: remove dependency on Docker CLI


AkihiroSuda:
The current CI uses both `nerdctl` and `docker` CLI.

The new release of `nerdctl` supports `nerdctl login` (thanks to @fahedouch), so dependency on Docker CLI should be now removable: https://github.com/AkihiroSuda/nerdctl/releases/tag/v0.6.0

ktock:
Thanks, I'll work on this.

PR #273, opened 2021-02-19
Use nerdctl instead of docker cli in the CI containers

ktock:
Fixes: #272

### DIFF
--- a/script/integration/containerd/entrypoint.sh
+++ b/script/integration/containerd/entrypoint.sh
@@ -23,8 +23,8 @@ set -euo pipefail
 source "/utils.sh"
 
 PLUGIN=stargz
-REGISTRY_HOST=registry-integration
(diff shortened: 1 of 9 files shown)


script/integration/containerd/entrypoint.sh
```diff
@@ -23,8 +23,8 @@ set -euo pipefail
 source "/utils.sh"
 
 PLUGIN=stargz
-REGISTRY_HOST=registry-integration
-REGISTRY_ALT_HOST=registry-alt
+REGISTRY_HOST=registry-integration.io
+REGISTRY_ALT_HOST=registry-alt.io
```
AkihiroSuda, line 27: I guess we need to fix nerdctl to accept the old hostname.

Changing the hostname is ok as a workaround, but I guess we should use .local rather than .io

AkihiroSuda (approved):
Thanks

PR #275, opened 2021-02-25
Support systemd notification

ktock:


go.mod
```diff
@@ -9,6 +9,7 @@ require (
 	github.com/containerd/continuity v0.0.0-20201208142359-180525291bb7
 	github.com/containerd/go-cni v1.0.1
 	github.com/containerd/stargz-snapshotter/estargz v0.4.1
+	github.com/coreos/go-systemd v0.0.0-20190321100706-95778dfbb74e
```
AkihiroSuda, line 12: We should use go-systemd/v22 module https://github.com/coreos/go-systemd/releases/tag/v22.1.0

AkihiroSuda (changes requested):

cmd/containerd-stargz-grpc/main.go
```diff
@@ -119,12 +121,31 @@ func main() {
...
+		notified, notifyErr := sddaemon.SdNotify(false, sddaemon.SdNotifyStopping)
+		log.G(ctx).Debugf("SdNotifyStopping notified=%v, err=%v", notified, notifyErr)
+	}
+	switch s {
+	case unix.SIGINT:
+		// do cleanup (defer) on SIGINT
+	default:
+		// exit
+		log.G(ctx).Info("Exiting")
+		os.Exit(0)
```
AkihiroSuda, line 143: Do we really need this?

ktock: We need this for avoiding normal overlayfs (non-lazy) layers from being removed during `*snapshot.Close()`.

go.mod
ktock: `k8s.io/apiserver`  depends on (`95778dfbb74e` = [v19?](https://github.com/coreos/go-systemd/releases/tag/v19)) and `go mod tidy` automatically downgrades to that version. 

https://github.com/kubernetes/apiserver/blob/217e5eef2f6e872b67c1a2a7150152962f341dba/go.mod#L11

AkihiroSuda: If you import v22 explicitly it wouldn't cause automatic downgrade

See https://github.com/containerd/containerd/blob/d8527867057daa93c17423ec7cd3438696b1c899/cmd/containerd/command/notify_linux.go#L24

cmd/containerd-stargz-grpc/main.go
AkihiroSuda: What does os.Exit do for that?

ktock: This avoids `defer` to be called (https://golang.org/pkg/os/#Exit).

go.mod
ktock: Thanks, fixed this.

cmd/containerd-stargz-grpc/main.go
AkihiroSuda: Ah, didn't know that. Please add comment lines for that. Or ideally we should disable the defer without depending on os.Exit.

AkihiroSuda (approved):

PR #281, opened 2021-03-05
PoC of "additional layer store" of Podman/CRI-O and lazy pulling of zstd:chunked

ktock:
This PR contains the following:

- PoC of "additional layer store" of Podman/CRI-O
- PoC of lazy pulling of zstd:chunked

[Zstd:chunked](https://github.com/containers/storage/pull/775) is a variant of zstd that enables extracting each file separately. The structure based on stargz and its verification logic is based on  eStargz. Currently, this is enabled by directly patching `estargz` lib for supporting that compression algorithm (zstd) and that footer format. Maybe we should make `estargz` lib configurable against compression algorithms and separate zstd:chunked patches into an indipendent package.

["Additional layer store"](https://github.com/containers/storage/pull/795) is a proposed functionality for https://github.com/containers/storage to provide remotely mounted layers to the runtime. Currently, this is implemented as a thin wrapper of our [`fs`](https://github.com/containerd/stargz-snapshotter/tree/master/fs) package. Instead of implementing snapshotter API, this directly exposes layers on the filesystem. This seems a bit out of scope as a non-core subproject of containerd. So this should be separated to another project, maybe?

The following command mounts the store at the `<mountpoint>`:

```
# registry-storage <mountpoint>
```

This store exposes the following structure required as an additional layer store:

```
<mountpoint>/base64(imageref)/<layerdigest>/
- layer : exposes the extracted layer
- info  : contains JSON-formatted "Layer" structure of containers/storage
- use   : files to notify the use of this layer (used for management of reference count)
```

There are also other APIs for debugging:

```
<mountpoint>/base64(imageref)/
- manifest       : contains manifest of this image
- config         : contains config blob of this image
- <layerdigest>/
  - chain        : exposes the extracted chain from the bottommost to this layer
```

### HelloBench result of Podman + {eStargz, zstd:chunked} on GitHub Actions (`Standard_DS2_v2` instance at `eastus2`):

https://github.com/ktock/stargz-snapshotter/actions/runs/623653929

I'll work on experiment on a richer instance with larger number of samples.

result

PR #293, opened 2021-03-26
Make estargz compression-algorithm-agnostic and support zstd (a.k.a. zstd:chunked)

ktock:
This is the subset of https://github.com/containerd/stargz-snapshotter/pull/281 but re-designed.

Initially, eStargz is based on gzip compression. But, through zstd:chunked work, it turned out that eStargz is not limited to gzip compression and the same chunking & verifying & prefetching method can be applied to other compression algorithms as well (e.g. zstd).

This commit makes `estargz` pkg configurable and agnostic about compression algorithms. For supporting non-gzip compression, the user must implement `estargz.Decompressor` and `estargz.Compressor` interfaces and must plug them to `estargz` tools (e.g. `estargz.Open` and `estargz.NewWriterWithCompression`). `estargz` also provides test suite that is usable for testing these non-gzip eStargz implementations.

This commit comes with `zstdchunked` pkg that support zstd compression for eStargz (a.k.a. zstd:chunked), based on the above extensibility. `zstdchunked` pkg contains `zstdchunked.Decompressor` and `zstdchunked.Compressor` that allows `estargz` pkg to use zstd compression (i.e. zstd:chunked) instead of gzip.

Layer converter and filesystem now support zstd:chunked leveraging `zstdchunked` pkg. `ctr-remote image optimize` and `ctr-remote image convert` support `--zstdchunked` option that omits zstd-based eStargz and filesystem supports zstd-based eStargz layers by default.

In the future, we can support other compression algorithms (e.g. lz4 used by nydus?) as well.

----

cc @giuseppe @AkihiroSuda


ktock:
@giuseppe WDYT? I know zstd:chunked is still under discussion (https://github.com/containers/storage/pull/775) and that spec may change in the future. But this implementation will help to make that discussion forward?

AkihiroSuda:
@giuseppe Could you take a look? 🙏

giuseppe:
sorry, I've missed the previous ping.

@ktock sure!  And we can try to make it as close as possible to the estargz TOC to simplify using it in stargz-snaphshotter.

The only difference I am aware of is how the chunking is represented.  Does the current metadata for files chunking work for you?

ktock:
@giuseppe Metadata (TOC) implemented in https://github.com/containers/storage/pull/775 looks good to me.
Could you add the following annotation to the result image for enabling chunk verification by stargz snapshotter?

- `containerd.io/snapshot/stargz/toc.digest` = uncompressed digest of manifest (TOC) (`sha256:deadbeef...`)

giuseppe:
> `containerd.io/snapshot/stargz/toc.digest` = uncompressed digest of manifest (TOC) (`sha256:deadbeef...`)

is it fine if it is named `io.containers.zstd-chunked.manifest-checksum` as it is currently?

ktock:
Containerd currently doesn't allow snapshotter to handle labels that are not prefixed by `containerd.io/snapshot/`. Additionally, `io.containers.zstd-chunked.manifest-checksum` seems stores "compressed" digest of TOC JSON but Stargz Snapshotter currently requires "uncompressed" digest of TOC JSON.

I also think we need standard spec for lazy-pulling-related annotations (currently discussed in https://github.com/opencontainers/image-spec/issues/815).

nativeconverter/zstdchunked/zstdchunked.go
```diff
@@ -0,0 +1,211 @@
...
+		uncompressedDesc := &desc
+		// We need to uncompress the archive first
+		if !uncompress.IsUncompressedType(desc.MediaType) {
+			var err error
+			uncompressedDesc, err = uncompress.LayerConvertFunc(ctx, cs, desc)
+			if err != nil {
+				return nil, err
+			}
+			if uncompressedDesc == nil {
+				return nil, errors.Errorf("unexpectedly got the same blob aftr compression (%s, %q)", desc.Digest, desc.MediaType)
```
AkihiroSuda, line 85: typo

nativeconverter/zstdchunked/zstdchunked.go
```diff
@@ -0,0 +1,211 @@
...
+		if err = w.Commit(ctx, n, "", content.WithLabels(labelz)); err != nil && !errdefs.IsAlreadyExists(err) {
+			return nil, err
+		}
+		if err := w.Close(); err != nil {
+			return nil, err
+		}
+		newDesc := desc
+		if uncompress.IsUncompressedType(newDesc.MediaType) {
+			if images.IsDockerType(newDesc.MediaType) {
+				newDesc.MediaType += ".zstd"
```
AkihiroSuda, line 173: Do we really want to support zstd for DockerType?

ktock: No, we don't. Fixed it to always convert to oci type.

AkihiroSuda (approved):

PR #331, opened 2021-06-02
Enable to embedded by k3s

ktock:
This change enables stargz snapshotter to embedded by k3s.
The following changes are included:

- Eliminate dependency to fusermount binary using go-fuse's `DirectMount` option
- Eliminate dependencies to the following packages specific to containerd 1.5 (k3s uses containerd 1.4.4)
  - `github.com/containerd/containerd/pkg/cri`
  - `github.com/containerd/containerd/snapshots/overlay/overlayutils`
- Support CRI-plugin-compatible registry configuration when builtin mode
  - `plugins."io.containerd.snapshotter.v1.stargz".registry` accepts the registry configuration same as `plugins."io.containerd.snapshotter.v1.cri".registry`.
  - This is only enabled for builtin plugin of stargz snapshotter. (`github.com/containerd/stargz-snapshotter/service/plugin`)
- Add tests for k3s using k3d


AkihiroSuda (approved):
LGTM, but how long will it take to update k3s to use containerd 1.5?

ktock:
> how long will it take to update k3s to use containerd 1.5?

I expect it to happen in k3s 1.22 (c.f. https://github.com/k3s-io/k3s/issues/3381).

PR #336, opened 2021-06-08
Enable to build containerd with the latest Makefile

ktock:
This fixes recent CI failures.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,7 +35,7 @@ RUN apt-get update -y && apt-get install -y libbtrfs-dev libseccomp-dev && \
     git clone -b ${CONTAINERD_VERSION} --depth 1 \
               https://github.com/containerd/containerd $GOPATH/src/github.com/containerd/containerd && \
     cd $GOPATH/src/github.com/containerd/containerd && \
-    GO111MODULE=off make && DESTDIR=/out/ make install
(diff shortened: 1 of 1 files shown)


AkihiroSuda (approved):

Dockerfile
```diff
@@ -35,7 +35,7 @@ RUN apt-get update -y && apt-get install -y libbtrfs-dev libseccomp-dev && \
     git clone -b ${CONTAINERD_VERSION} --depth 1 \
               https://github.com/containerd/containerd $GOPATH/src/github.com/containerd/containerd && \
     cd $GOPATH/src/github.com/containerd/containerd && \
-    GO111MODULE=off make && DESTDIR=/out/ make install
+    GO111MODULE=off make && DESTDIR=/out/ PREFIX= make install
```
thaJeztah, line 38: That feels odd (that it's now needed to set the PREFIX even when unused?)

thaJeztah: Ah, right; in that case, should it be

```suggestion
    GO111MODULE=off make && PREFIX=/out/ make install
```

Or is the intent to have binaries in `/out/` (not `/out/bin/`)?

ktock: Thank you for the suggestion! Yes, it looks better.

PR #324, opened 2021-05-19
[WIP] Add fuse-manager

ilyee:
## What is fusemanager

`fuse-manager` is a fuse daemon that handles all fuse processes, so fuse goroutines won't get killed when we restart `stargz-snapshotter`, and runnings containers don't need to be restarted.

By separating out `fuse-manager`, filesystem operations will all be handed by `fuse-manager`, and `stargz-snapshotter` is only responsable of snapshot operations.

## How to run it

1. Run `fuse-manager` with option `fusestore` (address for the fusemanager's store) and `socket` (address for the fusemanager's gRPC socket)
2. Run `stargz-snapshotter` with option `fusemanager-address` (same with `socket`), and `enable-fusemanager` with 'true' to enable it

## How it works

`fuse-manager` interacts with `stargz-snapshotter` through gRPC, its client implements `Filesystem` interface, so can be passed to `snbase.NewSnapshotter` to create a new snapshotter instance.

After running grpc server, `fuse-manager` waits for `Init` method to be called.  When `stargz-snapshotter` starts(or restarts), `Init` will be called to pass fs related configs to `fuse-manager`, which will create a new fs instance for handling new incoming `Mount`/`Check`/`Unmount` requests.

PR #341, opened 2021-06-18
Makefile: add `make generate` for proto code generation

ktock:
Currently, we don't have a common way to generate & validate proto-related codes. This commit adds this functionality to our CI leveraging `go generate` command.

This commit adds `make generate` command which generates proto codes. This runs `go generate -v ./...` for our repo in a containerized environment. The versions of tools (protoc and gogo) are enforced by that container. Contributors who edit proto files must run this command.

`make validate-generated` runs in our CI to check if all generated codes are expected. This runs `go generate -v ./...` again and checks if the generated files match to ones in the target commit.

### Example: generating proto codes for PR #324

The following file is needed as `fusemanager/api/generate.go`.

```golang
package api
//go:generate protoc --gogo_out=paths=source_relative,plugins=grpc:. api.proto
```

Then `make generate` generates `fusemanager/api/api.pb.go`.


script/generated-files/generate.sh
```diff
@@ -0,0 +1,89 @@
...
+      echo \${DIFFS} && \
+      exit 1 ; \
+    else \
+      echo OK ; \
+    fi
+EOF
+
+case ${COMMAND} in
+    update)
+        docker build --progress=plain --target update -f "${TMP_CONTEXT}/Dockerfile" -o - "${REPO}" | tar -C "${GENERATED_FILES}/" -x
```
AkihiroSuda, line 83: Set DOCKER_BUILDKIT explicitly

AkihiroSuda (approved):

Issue #366, opened 2021-07-12
Make GZIP compression levels configurable when optimizing an image


rdpsin:
Currently, there is no way for users to specify the GZIP compresssion level when optimizing their images to eStargz format, using the ctr-remote tool.

I propose a change that adds a new flag to the 'optimize' subcommand: estargz-compression-level (similar to the flag in convert subcommand) which defaults to 9. Users can specify compression level using --estargz-compression-level <value> while optimizing the image.

Also, changes to ctr-remote.md explaining the new flag and its usage.

The changes can be seen in this commit: https://github.com/rdpsin/stargz-snapshotter/commit/d69071baec43576653c4c14fd50a0e617f3a50cc

ktock:
@rdpsin SGTM. Would you like to open up the PR?

About the [proposed change]( https://github.com/rdpsin/stargz-snapshotter/blob/d69071baec43576653c4c14fd50a0e617f3a50cc/cmd/ctr-remote/commands/optimize.go#L145):

```golang
convertOpts = append(convertOpts, converter.WithLayerConvertFunc(logWrapper(f)), converter.WithLayerConvertFunc(estargzconvert.LayerConvertFunc(esgzOpts...)))
```

Rather than adding a new layer converter, could we have the something like the following function  (untested) in [nativeconverter pkg](https://github.com/containerd/stargz-snapshotter/blob/170bca44f2862599cbc209a8716ec3d008bf403c/nativeconverter/estargz/estargz.go) and use it in the [optimize.go ](https://github.com/containerd/stargz-snapshotter/blob/170bca44f2862599cbc209a8716ec3d008bf403c/cmd/ctr-remote/commands/optimize.go#L132)instead of `LayerConvertWithLayerOptsFunc`?

```golang
func LayerConvertWithLayerOptsAndCommonOptsFunc(opts map[digest.Digest][]estargz.Option, commonOpts ...estargz.Option) converter.ConvertFunc {
	if opts == nil {
		return LayerConvertFunc(commonOpts...)
	}
	return func(ctx context.Context, cs content.Store, desc ocispec.Descriptor) (*ocispec.Descriptor, error) {
		return LayerConvertFunc(append(commonOpts, opts[desc.Digest]...)...)(ctx, cs, desc)
	}
}
```

rdpsin:
Yes, that makes sense. I'll open a new PR soon.

ktock:
[...]

PR #368, opened 2021-07-13
Allow users to specify a GZIP compression level during optimize

rdpsin:
Currently, there is no way for users to specify the GZIP compresssion level when optimizing their images to eStargz format, using the ctr-remote tool.

This commit adds a new flag to the 'optimize' subcommand: estargz-compression-level (similar to the flag in convert subcommand) which defaults to 9. Users can specify compression level using --estargz-compression-level <value> while optimizing the image.

Also, changes to ctr-remote.md explaining the new flag and its usage.

Updated code with suggested changes in https://github.com/containerd/stargz-snapshotter/issues/366#issuecomment-878375257.

### DIFF
--- a/cmd/ctr-remote/commands/optimize.go
+++ b/cmd/ctr-remote/commands/optimize.go
@@ -17,6 +17,7 @@
 package commands
 
 import (
+	"compress/gzip"
 	"context"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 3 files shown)


ktock:
Thank you for updating the patch.

Linter seems not happy:

```
 cmd/ctr-remote/commands/optimize.go:82: File is not `gofmt`-ed with `-s` (gofmt)
			Name: "estargz-compression-level",
```

Please format the file.

rdpsin:
Done. Thanks.

ktock (approved):
Thanks!

PR #382, opened 2021-07-19
Enable GPUs and networking in container during optimize

rdpsin:
Currently, networking inside a container when optimizing through the ctr-remote tool is only
available through CNI-based networking plugins, not through `net-host`.  Furthermore, there is no option to enable GPUs
inside the container when optimizing GPU-based images.

This commit adds flags to the optimize command to enable both features.

Changes:

1) Added `net-host` flag to enable to networking inside the container, similar to when using `ctr-remote run`.
2) Added `gpus` flag to enable GPU-based image optimization.
3) Changes to documentation explaining the above made changes.

### DIFF
--- a/docs/ctr-remote.md
+++ b/docs/ctr-remote.md
@@ -66,6 +66,18 @@ The following example optimizes an image with a compression level of 1.
 # ctr-remote image optimize --oci --estargz-compression-level 1 ghcr.io/stargz-containers/golang:1.15.3-buster-org registry2:5000/golang:1.15.3-esgz
 ```
 
+You can enable host networking for the container using the `net-host` flag.
+
+```console
+# ctr-remote i optimize -t -i --oci --entrypoint='[ "/bin/bash", "-c" ]' --net-host --args='[ "ip a && curl example.com" ]' ghcr.io/stargz-containers/centos:8-test registry2:5000/centos:8-test-esgz
+```
+
+You can optimize GPU-based images using the `gpu` flag. The flag expects a comma separated list of integers.
+
+```console
+# ctr-remote i optimize --oci --gpus "0" <src> <target>
+```
+
 `--oci` option is highly recommended to add when you create eStargz image.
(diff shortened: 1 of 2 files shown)


docs/ctr-remote.md
```diff
@@ -66,6 +66,18 @@ The following example optimizes an image with a compression level of 1.
 # ctr-remote image optimize --oci --estargz-compression-level 1 ghcr.io/stargz-containers/golang:1.15.3-buster-org registry2:5000/golang:1.15.3-esgz
 ```
 
+You can enable networking inside the container using the `net-host` flag.
```
ktock, line 69: Networking inside container is possible with CNI as well.
e.g. you can use `--cni` option in the demo container([Dockerfile](https://github.com/containerd/stargz-snapshotter/blob/c96a7f9661d70eda53f3f66c09cdbdafcf4ca235/Dockerfile#L187-L197), [compose](https://github.com/containerd/stargz-snapshotter/blob/c96a7f9661d70eda53f3f66c09cdbdafcf4ca235/script/demo/docker-compose.yml)). [`ctr-remote i optimize` is also tested in the container](https://github.com/containerd/stargz-snapshotter/blob/c96a7f9661d70eda53f3f66c09cdbdafcf4ca235/script/optimize/optimize/entrypoint.sh#L260-L331) in our CI.

rdpsin: Thank you for your feedback. I'm aware of using CNI plugins to enable networking. I suggested the `net-host` flag as a simpler way to enable networking (similar to `ctr run`) inside the container, since setting up CNI can be more troublesome/error-prone, especially if all you want is basic network connectivity.

ktock: So I think rather than "You can enable networking inside the container ",  "You can enable *host* networking for the container` seems the right description here.

rdpsin: Makes sense. I'll update the docs.

ktock (approved):
LGTM

docs/ctr-remote.md
ktock: Thank you!

PR #389, opened 2021-07-27
Update stargz-snapshotter.service

yylt:
fix correct service unit


script/config/etc/systemd/system/stargz-snapshotter.service
```diff
@@ -8,7 +8,7 @@ Type=notify
 Environment=HOME=/root
 ExecStart=/usr/local/bin/containerd-stargz-grpc --log-level=debug --config=/etc/containerd-stargz-grpc/config.toml
 Restart=always
-RestartSec=1
+RestartSec=1s
```
ktock, line 11: Unit-less second is a valid value.

https://www.freedesktop.org/software/systemd/man/systemd.service.html#RestartSec=

> Takes a unit-less value in seconds,

yylt: thanks

PR #394, opened 2021-08-04
Add log lines and update log levels to aid debugging

iamsumee:
This patch adds a few new log lines. It also updates log levels of some
error cases from debug to warn.


ktock (approved):
LGTM. Thank you!

PR #407, opened 2021-08-09
Make fuse timeout values configurable

iamsumee:
stargz-snapshotter currently sets fuse timeout values to one second. Container
images are immutable and hence some applications may want to set higher
timeout values.

This patch makes the timeout values configurable.

### DIFF
--- a/fs/config/config.go
+++ b/fs/config/config.go
@@ -52,6 +52,8 @@ type Config struct {
 
 	// DirectoryCacheConfig is config for directory-based cache.
 	DirectoryCacheConfig `toml:"directory_cache"`
+
+	FuseConfig `toml:"fuse"`
 }
 
 type BlobConfig struct {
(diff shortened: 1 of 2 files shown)


ktock:
Please fix linter error (go fmt might be needed for fs/fs.go)

https://github.com/containerd/stargz-snapshotter/pull/407/checks?check_run_id=3285647523#step:3:123

```
fs/fs.go:68: File is not `gofmt`-ed with `-s` (gofmt)
	defaultFuseTimeout = time.Second
```

fs/config/config.go
```diff
@@ -68,3 +70,13 @@ type DirectoryCacheConfig struct {
 	SyncAdd          bool `toml:"sync_add"`
 	Direct           bool `toml:"direct"`
 }
+
+type FuseConfig struct {
+	// This defines overall timeout attribute for a file system in seconds.
+	//
```
ktock, line 76: Could you replace `this` to `AttrTimeout`? And the empty line isn't needed.

fs/config/config.go
```diff
@@ -68,3 +70,13 @@ type DirectoryCacheConfig struct {
...
+	AttrTimeout int64 `toml:"attr_timeout"`
+
+	// This defines TTL for directory, name lookup in seconds.
+	//
```
ktock, line 80: Could you replace `this` to `EntryTimeout`? And the empty line isn't needed.

fs/config/config.go
iamsumee: Ack

fs/config/config.go
iamsumee: Ack

ktock (approved):

PR #408, opened 2021-08-09
Add image launch metrics

vkuzniet:
This PR adds the metrics for container image launch.
The list of metrics is as follows: 

- Prefetch bytes per layer 
- GET latency for stargz header 
- GET latency for stargz footer 
- GET latency for stargz TOC 
- Latency to deserialize json TOC 
- Total time container launch is blocked on prefetch complete

Signed-off-by: Viktor Kuznietsov <vkuzniet@amazon.com>

### DIFF
--- a/estargz/estargz_test.go
+++ b/estargz/estargz_test.go
@@ -107,6 +107,202 @@ func legacyFooterBytes(tocOff int64) []byte {
 	return buf.Bytes()
 }
 
+func TestWriteAndOpenWithTelemetry(t *testing.T) {
+	const content = "Some contents"
+	invalidUtf8 := "\xff\xfe\xfd"
+
+	xAttrFile := xAttr{"foo": "bar", "invalid-utf8": invalidUtf8}
+
+	tests := []struct {
+		name      string
+		chunkSize int
+		in        []tarEntry
+		want      []stargzCheck
+		wantNumGz int        // expected number of gzip streams
+		telemetry *Telemetry // telemetry hooks configuration
+	}{
+		{
+			name: "no_telemetry",
+			in: tarOf(
+				dir("foo/"),
+				file("foo/bar.txt", content, xAttrFile),
+			),
+			wantNumGz: 4, // var dir, foo.txt alone, TOC, footer
+			want: checks(
+				numTOCEntries(2),
+				hasDir("foo/"),
+				hasFileLen("foo/bar.txt", len(content)),
+				hasFileDigest("foo/bar.txt", digestFor(content)),
+				hasFileContentsRange("foo/bar.txt", 0, content),
+				hasFileContentsRange("foo/bar.txt", 1, content[1:]),
+				entryHasChildren("", "foo"),
+				entryHasChildren("foo", "bar.txt"),
+				hasFileXattrs("foo/bar.txt", "foo", "bar"),
+				hasFileXattrs("foo/bar.txt", "invalid-utf8", invalidUtf8),
+			),
+			telemetry: nil,
+		},
+		{
+			name: "no_get_footer",
+			in: tarOf(
+				dir("foo/"),
+				file("foo/bar.txt", content, xAttrFile),
+			),
+			wantNumGz: 4, // var dir, foo.txt alone, TOC, footer
+			want: checks(
+				numTOCEntries(2),
+				hasDir("foo/"),
+				hasFileLen("foo/bar.txt", len(content)),
+				hasFileDigest("foo/bar.txt", digestFor(content)),
+				hasFileContentsRange("foo/bar.txt", 0, content),
+				hasFileContentsRange("foo/bar.txt", 1, content[1:]),
+				entryHasChildren("", "foo"),
+				entryHasChildren("foo", "bar.txt"),
+				hasFileXattrs("foo/bar.txt", "foo", "bar"),
+				hasFileXattrs("foo/bar.txt", "invalid-utf8", invalidUtf8),
+			),
+			telemetry: &Telemetry{
+				GetTocLatency: func(t time.Time) {
+
+				},
+				DeserializeTocLatency: func(t time.Time) {
+
+				},
+			},
+		},
+		{
+			name: "no_get_toc",
+			in: tarOf(
+				dir("foo/"),
+				file("foo/bar.txt", content, xAttrFile),
+			),
+			wantNumGz: 4, // var dir, foo.txt alone, TOC, footer
+			want: checks(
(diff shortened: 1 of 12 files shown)


vkuzniet:
Hi @ktock,
It seems that `estargz` is a separate module which brings a challenge with metrics infrastructure. The `stargz snapshotter` builds well, however if you go with just `estargz` part, it does not see any metrics from the snapshotter for obvious reasons.
Can you help me learn why `estargz` is a separate module and what would be the preferred way to add metrics there given it's special nature?
My thoughts were to simply add a separate `/metrics` package to `estargz` and extend it accordingly.
Let me know what do you think?

ktock:
@vkuzniet  Thank you for this PR.

> Can you help me learn why estargz is a separate module 

estargz builder & parser are used by projects outside of stargz-snapshotter. They don't require most snapshotter-related dependencies (e.g. FUSE, containerd, etc) so we have a minimal module `estargz`.

> GET latency for stargz header
> GET latency for stargz footer
> GET latency for stargz TOC
> Latency to deserialize json TOC

The performance of each parsing step of `estargz.Open` can be analyzed using other methods like go pprof. Isn't it enough to just emit the latency of calling `estargz.Open`?

vkuzniet:
Hi @ktock!

> estargz builder & parser are used by projects outside of stargz-snapshotter. They don't require most snapshotter-related dependencies (e.g. FUSE, containerd, etc) so we have a minimal module `estargz`.

Thank you very much for clarification.


> > GET latency for stargz header
> > GET latency for stargz footer
> > GET latency for stargz TOC
> > Latency to deserialize json TOC
> 
> The performance of each parsing step of `estargz.Open` can be analyzed using other methods like go pprof. Isn't it enough to just emit the latency of calling `estargz.Open`?

I agree that you can analyze the performance on a single host level with `go pprof`, but this will not work in production environment. Given that `estargz.Open` has multiple components to it, it'd be good to have visibility into which component is taking the most time.

ktock:
@vkuzniet

> I agree that you can analyze the performance on a single host level with go pprof, but this will not work in production environment. Given that estargz.Open has multiple components to it, it'd be good to have visibility into which component is taking the most time.

Sure. How about adding an option for hooks (e.g. `estargz.WithXXXHook`) to `estargz.Open` func (maybe we need a new func `estargz.OpenWithOptions` for keeping API copmatibility) for allowing the client to inject a hook for each parse step?
Then we can inject metrics logic in `fs/reader` pkg leveraging this inteace.

This allows us to have metrics without adding metrics dependencies directly to `estargz`.

vkuzniet:
@ktock 

> Sure. How about adding an option for hooks (e.g. `estargz.WithXXXHook`) to `estargz.Open` func (maybe we need a new func `estargz.OpenWithOptions` for keeping API copmatibility) for allowing the client to inject a hook for each parse step?
> Then we can inject metrics logic in `fs/reader` pkg leveraging this inteace.
> 
> This allows us to have metrics without adding metrics dependencies directly to `estargz`.

Sounds like a plan. Will add it to the next revision of my PR.

estargz/estargz_test.go
```diff
@@ -107,6 +107,358 @@ func legacyFooterBytes(tocOff int64) []byte {
 	return buf.Bytes()
 }
 
+func TestWriteAndOpenWithTelemetry_AllHooksDefined(t *testing.T) {
```
ktock, line 110: Could we combine `TestWriteAndOpenWithTelemetry_*` into a single test function?

fs/reader/reader.go
```diff
@@ -92,8 +94,20 @@ func (nv nopVerifier) Verified() bool {
 // NewReader creates a Reader based on the given stargz blob and cache implementation.
 // It returns VerifiableReader so the caller must provide a estargz.TOCEntryVerifier
 // to use for verifying file or chunk contained in this stargz blob.
-func NewReader(sr *io.SectionReader, cache cache.BlobCache) (*VerifiableReader, error) {
-	r, err := estargz.Open(sr)
+func NewReader(sr *io.SectionReader, cache cache.BlobCache, layerDigest digest.Digest) (*VerifiableReader, error) {
```
ktock, line 97: Could we pass `*estargz.Telemetry` to `NewReader()` instead of `digest.Digest`? It would be great if we can reuse this `*estargz.Telemetry` also in `reader.Cache()`.

snapshot/snapshot.go
```diff
@@ -265,7 +268,11 @@ func (o *snapshotter) Prepare(ctx context.Context, key, parent string, opts ...s
 		}
 	}
 
-	return o.mounts(ctx, s, parent)
+	start := time.Now()
+	dgst := digest.FromString("") // we pass the digest as an empty string here, since the metric we capture is layer agnostic
+	mounts, err := o.mounts(ctx, s, parent)
+	commonmetrics.MeasureLatency(commonmetrics.PrefetchesCompleted, dgst, start) // measuring the time the container launch is blocked on prefetch to complete
```
ktock, line 274: Why not measure this in [`layer.WaitForPrefetchCompletion()`](https://github.com/containerd/stargz-snapshotter/blob/v0.7.0/fs/layer/layer.go#L90-L91)?

vkuzniet: > Why not measure this in [`layer.WaitForPrefetchCompletion()`](https://github.com/containerd/stargz-snapshotter/blob/v0.7.0/fs/layer/layer.go#L90-L91)?

We can't do that because it is only invoked if prefetch is enabled. Will you be OK with measuring it `Check` function defined in `fs.go`?

fs/reader/reader.go
vkuzniet: That works. Will update in the next revision of the PR.

estargz/estargz_test.go
vkuzniet: Yes, will update in the next revision of the PR.

snapshot/snapshot.go
ktock: > ill you be OK with measuring it in Check function defined in fs.go?

SGTM :)

estargz/estargz_test.go
vkuzniet: All set.

fs/reader/reader.go
vkuzniet: All set.

snapshot/snapshot.go
vkuzniet: Updated.

ktock (approved):
Thank you!

PR #409, opened 2021-08-11
Allow users to configure prefetch chunk size

iamsumee:
stargz-snapshotter fetches entire prefetch range using a single HTTP GET request.
This reduces number of requests to remote registry. However some registries give
better throughput with multiple concurrent requests rather than a single GET request.

This patch adds a configurable parameter to enable users to decide the optimal
prefetch chunk size for their registries.

### DIFF
--- a/fs/remote/blob.go
+++ b/fs/remote/blob.go
@@ -36,6 +36,7 @@ import (
 	"github.com/containerd/stargz-snapshotter/fs/source"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
+	"golang.org/x/sync/errgroup"
 )
 
 var contentRangeRegexp = regexp.MustCompile(`bytes ([0-9]+)-([0-9]+)/([0-9]+|\\*)`)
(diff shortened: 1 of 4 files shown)


fs/remote/blob.go
```diff
@@ -151,6 +153,24 @@ func (b *blob) FetchedSize() int64 {
 	return sz
 }
 
+func (b *blob) fetch(offset int64, size int64, fr *fetcher, cacheOpts *options) error {
```
ktock, line 156: The function name should be `cache`.

fs/remote/blob.go
```diff
@@ -165,20 +185,31 @@ func (b *blob) Cache(offset int64, size int64, opts ...Option) error {
...
+				return b.fetch(start, end-start, fr, &cacheOpts)
+			}
+			return b.fetch(start, fetchSize, fr, &cacheOpts)
+		})
+
+		begin = begin + fetchSize
 	}
 
-	return b.fetchRange(discard, &cacheOpts)
+	return eg.Wait()
```
ktock, line 212: Could it be simplified like the following?

https://play.golang.org/p/4k-6WwbzlXy

```golang
	end := offset + size
	for i := offset; i < end ; i += fetchSize {
		i, l := i, fetchSize
		if i + l > end {
			l = end - i
		}
		eg.Go(func() error {
			return b.fetch(i, l, fr, &cacheOpts)
		})
	}
	return eg.Wait()
```

fs/remote/blob.go
iamsumee: `blob` has a field named `cache` too. 

The build fails with following error:

```
fs/remote/blob.go:156:6: type blob has both field and method named cache
make: *** [containerd-stargz-grpc] Error 2
```

fs/remote/blob.go
iamsumee: Ack. I'll update and send out another iteration.

ktock (approved):
Thank you!

PR #415, opened 2021-08-18
Store filesystem metadata on disk

ktock:
Currently, stargz snapshotter holds filesystem metadata in memory but this ends up consuming a large amount of memory and they aren't free until the filesystem layer is unmounted. This commit tries to solve this problem by fixing snapshotter to store the metadata on disk using bbolt.

## Performance comparison

Summary:

- PR improves memory consumption (about 5x smaller)
- PR increases `*node.Getattr` latency (about 2x ~ 3x slower) but actual workloads (HelloBench, stat commands) don't report significant performance degradation
  - Welcome for any suggestions about benchmarks where this change will be problematic

### Memory consumption

- host memory: ~16GB
- command: `nerdctl --snapshotter=stargz --insecure-registry run -it --rm registry2:5000/kdeneon/plasma:unstable-esgz echo hello`
  - `registry2` runs as a container, on the same host of contianerd.
  - size of `registry2:5000/kdeneon/plasma:unstable-esgz ` is 1.1 GiB according to containerd.
- config
  ```
  metrics_address = "127.0.0.1:8234"
  disable_verification = false
  [[resolver.host."registry2:5000".mirrors]]
  host = "registry2:5000"
  insecure = true
  [directory_cache]
  direct = true
  ```

This change makes memory consumption 5x smaller.

|main|PR|
|---|---|
|main-mem02|entriesdb-mem02|

#### latency of `*node.Getattr` operation (95 pctl.)

- image: ghcr.io/stargz-containers/postgres:13.1-esgz
- command: `time nerdctl exec -it $NAME find /usr/ -type d -exec /bin/bash -c "stat {}/*" \;`
- config: same as the above

This change makes `*node.Getattr` operation about 2x ~ 3x slower.
Though this doesn't seems to cause large performance degradation in actual workloads (HelloBench and stat commands) as shown in the following sections, we should continuously seek the way to optimize the performance of reading metadata.

|main|PR|
|---|---|
|main-getattr01|entriesdb-getattr01|

Total time for `find /usr/ -type d -exec /bin/bash -c "stat {}/*" \;`:
(Reading metadata doesn't seem to be a bottleneck of this command?)

|main|PR|
|---|---|
|10.209s|10.112s|

### HelloBench

https://github.com/ktock/stargz-snapshotter/runs/3351635146

result


ktock:
Moved telemetry hooks from `estargz.Reader` to `metadata.Reader`. `stargz_toc_json_deserialize` is deleted because we now perform JSON decoding per TOCEntry not per TOC JSON. cc @vkuzniet

iamsumee:
Intuitively I expect estargz-noopt images to have higher overall execution time than estargz. Benchmarks with tomcat image suggests otherwise. Do we understand why is estargz image worse off than estargz-noopt?

ktock:
@iamsumee

> Given that Getattr latencies increase with this patch, can we make these changes as an opt-in feature?

Sure, I'll add a configration field for this in `config.toml`

>    Intuitively I expect estargz-noopt images to have higher overall execution time than estargz. Benchmarks with tomcat image suggests otherwise. Do we understand why is estargz image worse off than estargz-noopt?
>    Can we please benchmark with https://hub.docker.com/_/rabbitmq, https://hub.docker.com/_/jenkins and elastic search images?

I'll benchmark it with more num of samples to investigate this deeper.
I'll add the result of  rabbitmq, jenkins and elastic search as well.

> What were the vCPU and network bandwidths on the host where the benchmarks were run?

- HelloBench: mesured on Github Actions ubuntu runner ([spec](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#cloud-hosts-for-github-hosted-runners)) + ghcr.io : https://github.com/ktock/stargz-snapshotter/runs/3351635146
- other: internally
  - cores: 8
  - registry serves as a local container on the same host as containerd ([compose](https://github.com/containerd/stargz-snapshotter/blob/main/script/demo)).
    - bandwidth: wget shows 1.74 GB/s avg.
      ```
      $ head -c 10G </dev/urandom > ~/bigblob
      $ DIGEST=$(cat ~/bigblob | sha256sum | sed -E 's/([^ ]*).*/sha256:\1/g')
      $ UPLOAD_URL=$(curl -siL -X POST "registry2:5000/v2/test/blobs/uploads/" | grep "Location:" | cut -d " " -f 2 | tr -d "[:space:]")
      $ curl -X PUT -H "Content-Type: application/octet-stream" --data-binary "@${HOME}/bigblob" "${UPLOAD_URL}&digest=${DIGEST}"
      $ for i in {1..3}; do wget -O /dev/null --show-progress registry2:5000/v2/test/blobs/${DIGEST} ; done
      /dev/null                             100%[========================================================================>]  10.00G  1.85GB/s    in 5.4s
      ...
      /dev/null                             100%[========================================================================>]  10.00G  1.83GB/s    in 5.6s
      ...
      /dev/null                             100%[========================================================================>]  10.00G  1.56GB/s    in 6.4s   
      ```

fs/layer/layer.go
```diff
@@ -157,7 +160,16 @@ func NewResolver(root string, backgroundTaskManager *task.BackgroundTaskManager,
 		logrus.WithField("key", key).Debugf("cleaned up blob")
 	}
 
+	if err := os.MkdirAll(root, 0700); err != nil {
+		return nil, err
+	}
+	db, err := bolt.Open(filepath.Join(root, "metadata.db"), 0666, nil)
```
kzys, line 166: Does it have to be writable from all users? While MkdirAll above would restrict who can access the file, it would be better to narrow down the permissions if possible.

fs/layer/node_test.go
```diff
@@ -307,8 +317,25 @@ func TestExistence(t *testing.T) {
...
+func newReader(sr *io.SectionReader) (_ *metadata.Reader, closeFn func(), _ error) {
+	f, err := ioutil.TempFile("", "layertestdb")
+	if err != nil {
+		return nil, nil, err
+	}
+	db, err := bolt.Open(f.Name(), 0666, nil)
+	if err != nil {
+		return nil, nil, err
+	}
+	closeFn = func() { db.Close(); os.Remove(f.Name()); f.Close() }
```
kzys, line 329: You could call f.Close() right before bolt.Open() since bolt.Open anyway opens a new file.

store/pool.go
```diff
@@ -120,6 +127,22 @@ type Pool struct {
...
+
+	idMap   map[uint32]struct{}
+	idMapMu sync.Mutex
+}
+
+func (p *Pool) uniqueID() (id uint32, _ error) {
+	p.idMapMu.Lock()
+	defer p.idMapMu.Unlock()
+	for i := 0; i < 10000; i++ {
+		id = rand.New(rand.NewSource(time.Now().UnixNano())).Uint32()
```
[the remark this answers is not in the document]
kzys: +1 How about using rand.Unit32() here? We can seed Go's default random number generator in the main function.

rdpsin:
Hello, 

I did some benchmarking. Here are the results.

**Pull time:** Time to execute `ctr-remote i rpull $IMAGE`
**Run time:** Time to execute `ctr-remote run --net-host $IMAGE`

**w/o patch**

Image | Pull time (ms) | Run time (ms)
------|-------------|--------
busybox | 604 | 214
rabbitmq | 1155 | 16170
ghost | 3498 | 9206
drupal | 2775 | 716

**w/ patch**

Image | Pull time (ms) | Run time (ms)
------|-------------|--------
busybox | 662 | 512
rabbitmq | 845 | 22265
ghost | 1888 | 12978
drupal | 2111 | 1242

Benchmarks were done on a 4-core machine. 

Memory consumption shows a definite improvement (upto 85%).

rdpsin:
I tried to find the root cause for the runtime regression. Profiling shows `cacheWithReader` taking on average thrice as long as before. (However, that includes outliers where `cacheWithReader` latencies would spike significantly).


A few example of latencies:

Image: Python (https://hub.docker.com/_/python) estargz gzip max compression

(Latencies are in nanoseconds)

**w/o patch**

```
cacheWithReader time=389
cacheWithReader time=510
cacheWithReader time=12471
cacheWithReader time=507
cacheWithReader time=607
cacheWithReader time=6710
cacheWithReader time=1113
cacheWithReader time=6204
cacheWithReader time=366
cacheWithReader time=393
cacheWithReader time=242
cacheWithReader time=5745
cacheWithReader time=147
cacheWithReader time=858
```

**w/ patch**
```
cacheWithReader time=47216
cacheWithReader time=374436
cacheWithReader time=29083
cacheWithReader time=24296
cacheWithReader time=93903
cacheWithReader time=8407
cacheWithReader time=167736
cacheWithReader time=6486
cacheWithReader time=78370
cacheWithReader time=130138
cacheWithReader time=159985
cacheWithReader time=403028
cacheWithReader time=6308
```

rdpsin:
Can we change all `db.Update` calls to `db.Batch` and make `MaxBatchDelaySize` and `MaxBatchDelayTime` configurable? There is quite a bit of performance improvement when we do that.

**w/ Update**

Image | Pull time (ms) | Run time (ms) |  Total time (ms)
-------|---------------|---------------|-------------------
busybox | 662 | 512 | 1174
rabbitmq | 845 | 22265 | 23110
ghost | 1888 | 12978 | 14867
drupal | 2111 | 1242 | 3353

**w/ Batch (MaxBatchDelaySize = 1000, MaxBatchDelayTime = 10ms)**

and 

```
bbolt.Options = {
      NoFreelistSync: true,
      InitialMmapSize: 64M,
      FreelistType: FreelistMapType,
}
```

Image | Pull time (ms) | Run time (ms) |  Total time (ms)
-------|---------------|---------------|-------------------
busybox | 569 | 249 | 818
rabbitmq | 924 | 17291 | 18215
ghost | 1324 | 9529 | 10853
drupal | 1463 | 946 | 2409

ktock:
Thank you for the suggestion! I'll try the options and methods.
btw I'm curious what makes the `run` performance of `rabbitmq` slower (seems to happen even without this patch).

PR #419, opened 2021-08-19
Added helpful message whenever optimize period expires

rdpsin:
Added a message that will print after the period to monitor access logs times out. Currently, the message displayed provides no indication why the task was killed.


ktock (approved):
Thank you!

PR #427, opened 2021-08-23
Upgrade golangci-lint from v1.39.0 to v1.42.0

kzys:
The first commit is fixing Makefile. The second commit is actually upgrading golangci-lint.


ktock (approved):
Thank you!

PR #429, opened 2021-08-24
Add /debug/ endpoints to expose diagnostic information

kzys:
As like containerd, this change adds a few /debug/ endpoints. They can
be used from `ctr pprof` command.

### DIFF
--- a/cmd/containerd-stargz-grpc/main.go
+++ b/cmd/containerd-stargz-grpc/main.go
@@ -34,6 +34,7 @@ import (
 	"github.com/containerd/containerd/log"
 	"github.com/containerd/containerd/pkg/dialer"
 	"github.com/containerd/containerd/snapshots"
+	"github.com/containerd/containerd/sys"
 	"github.com/containerd/stargz-snapshotter/service"
 	"github.com/containerd/stargz-snapshotter/service/keychain/cri"
 	"github.com/containerd/stargz-snapshotter/service/keychain/dockerconfig"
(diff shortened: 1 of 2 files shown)


cmd/containerd-stargz-grpc/server.go
```diff
@@ -0,0 +1,34 @@
...
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package main
+
+import (
```
kzys, line 19: Since main.go imports quite a lot, I'd like to have a separate file for this feature.

cmd/containerd-stargz-grpc/main.go
```diff
@@ -75,6 +76,9 @@ type snapshotterConfig struct {
 
 	// NoPrometheus is a flag to disable the emission of the metrics
 	NoPrometheus bool `toml:"no_prometheus"`
+
+	// DebugAddress is a Unix domain socket address where the snapshotter exposes /debug/ endpoints.
+	DebugAddress string `toml:"debug_address"`
```
kzys, line 81: Right now `debug` is already taken by a boolean flag. Does it make sense to have `debug` section as like containerd? It would be a breaking change though.

ktock: We don't have many debug options as of now so having a flat field (debug_address) looks good to me. If we have too many debug options in the future, we can consider reorganizing the configuration fields (e.g. introducing debug section).

ktock (approved):
Thank you!

PR #433, opened 2021-08-29
store: Release resources when no reference to a layer

ktock:
This adds GC functionality to stargz store when there is no reference to a layer.
Stargz store already has reference counter.
This commit adds cleanup codes executed when the counter becomes 0.


AkihiroSuda (approved):

AkihiroSuda:
Would be good to have tests for GC, but can be another PR

PR #441, opened 2021-09-01
Reduce required environment variables on "make benchmark"

kzys:
"make benchmark" needs multiple environment variables to be set
correctly.

~This change makes some of them optional. The rest are~

The only remaining environment variable is

- BENCHMARK_TARGETS
- ~BENCHMARK_RUNTIME_MODE~
- ~DOCKER_BUILDKIT~

### DIFF
--- a/script/benchmark/tools/percentiles.sh
+++ b/script/benchmark/tools/percentiles.sh
@@ -40,9 +40,9 @@ if [ ${#IMAGES[@]} -eq 0 ] ; then
     IMAGES=( $(cat "${JSON}" | jq -r '[ .[] | select(.mode=="'${MODES[0]}'").repo ] | unique[]') )
 fi
 
-GRANULARITY="${BENCHMARK_PERCENTILES_GRANULARITY}"
+GRANULARITY="${BENCHMARK_PERCENTILES_GRANULARITY:-}"
 if [ "${GRANULARITY}" == "" ] ; then
-    GRANULARITY="0.1"
+    GRANULARITY="10"
 fi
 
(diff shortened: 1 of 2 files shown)


kzys:
Does it make sense to set DOCKER_BUILDKIT=1 and BENCHMARK_RUNTIME_MODE=containerd?

The first variable apparently changes Docker slightly in addition to speed up. I couldn't build images without BuildKit. The second is, well, I'm biased :)

script/benchmark/tools/percentiles.sh
```diff
@@ -40,7 +40,7 @@ if [ ${#IMAGES[@]} -eq 0 ] ; then
     IMAGES=( $(cat "${JSON}" | jq -r '[ .[] | select(.mode=="'${MODES[0]}'").repo ] | unique[]') )
 fi
 
-GRANULARITY="${BENCHMARK_PERCENTILES_GRANULARITY}"
+GRANULARITY="${BENCHMARK_PERCENTILES_GRANULARITY:-}"
 if [ "${GRANULARITY}" == "" ] ; then
     GRANULARITY="0.1"
```
kzys, line 45: Hmm, 0.1 seems too slow

ktock: We can change it (to `1` or `10`?)

ktock:
> Does it make sense to set DOCKER_BUILDKIT=1 and BENCHMARK_RUNTIME_MODE=containerd by default?

SGTM :)

ktock (approved):
Thank you!

PR #445, opened 2021-09-01
Run ctr pprof against containerd-stargz-grpc in Hello Bench

kzys:
containerd-stargz-grpc exposes its debug endpoint since e60759e.
Now Hello Bench in "make benchmark" can run "ctr pprof" against
the endpoint.


ktock (approved):
LGTM

PR #446, opened 2021-09-02
Avoid downloading the same data from the registry multiple times

vkuzniet:
stargz snapshotter does not have the mechanism to prevent downloading the same data (or overlapping data) from the registry multiple times. This leads to waste of network bandwidth as it ends up fetching the same bytes multiple times.
Following log lines record the fact that this is happening (the logs were added to `fetchRange` function in `fs/remote/blob.go` and are not the part of this PR):
`
time="2021-08-09T22:33:27.134794144Z" level=debug msg="request header=map[Accept-Encoding:[identity] Range:[bytes=167772160-168820735]] layer=sha256:8ea1027575e92df4962031bccc31de8d8ee052960a50b7649a1f964c50bef79c."
time="2021-08-09T22:33:27.135094521Z" level=debug msg="request header=map[Accept-Encoding:[identity] Range:[bytes=167772160-168820735]] layer=sha256:8ea1027575e92df4962031bccc31de8d8ee052960a50b7649a1f964c50bef79c."
time="2021-08-09T22:33:27.135910229Z" level=debug msg="request header=map[Accept-Encoding:[identity] Range:[bytes=167772160-168820735]] layer=sha256:8ea1027575e92df4962031bccc31de8d8ee052960a50b7649a1f964c50bef79c."
`

This PR addresses this issue.

Signed-off-by: Viktor Kuznietsov <vkuzniet@amazon.com>

### DIFF
--- a/fs/remote/blob.go
+++ b/fs/remote/blob.go
@@ -28,6 +28,8 @@ import (
 	"io"
 	"io/ioutil"
 	"regexp"
+	"sort"
+	"strings"
 	"sync"
 	"time"
 
@@ -37,6 +39,7 @@ import (
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
 	"golang.org/x/sync/errgroup"
+	"golang.org/x/sync/singleflight"
 )
 
 var contentRangeRegexp = regexp.MustCompile(`bytes ([0-9]+)-([0-9]+)/([0-9]+|\\*)`)
@@ -64,13 +67,31 @@ type blob struct {
 	checkInterval     time.Duration
 	fetchTimeout      time.Duration
 
-	fetchedRegionSet   regionSet
-	fetchedRegionSetMu sync.Mutex
+	fetchedRegionSet    regionSet
+	fetchedRegionSetMu  sync.Mutex
+	fetchedRegionGroup  singleflight.Group
+	fetchedRegionCopyMu sync.Mutex
 
 	resolver *Resolver
 
 	closed   bool
 	closedMu sync.Mutex
+}
+
+func makeBlob(fetcher *fetcher, size int64, chunkSize int64, prefetchChunkSize int64,
+	blobCache cache.BlobCache, lastCheck time.Time, checkInterval time.Duration,
+	r *Resolver, fetchTimeout time.Duration) *blob {
+	return &blob{
+		fetcher:           fetcher,
+		size:              size,
+		chunkSize:         chunkSize,
+		prefetchChunkSize: prefetchChunkSize,
+		cache:             blobCache,
+		lastCheck:         lastCheck,
+		checkInterval:     checkInterval,
+		resolver:          r,
+		fetchTimeout:      fetchTimeout,
+	}
 }
 
 func (b *blob) Close() error {
@@ -151,6 +172,17 @@ func (b *blob) FetchedSize() int64 {
 	sz := b.fetchedRegionSet.totalSize()
 	b.fetchedRegionSetMu.Unlock()
 	return sz
+}
+
+func makeSyncKey(allData map[region]io.Writer) string {
+	keys := make([]string, len(allData))
+	keysIndex := 0
+	for key := range allData {
+		keys[keysIndex] = fmt.Sprintf("[%d,%d]", key.b, key.e)
+		keysIndex++
+	}
+	sort.Strings(keys)
+	return strings.Join(keys, ",")
 }
 
 func (b *blob) cacheAt(offset int64, size int64, fr *fetcher, cacheOpts *options) error {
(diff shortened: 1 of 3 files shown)


fs/remote/blob.go
```diff
@@ -67,12 +67,37 @@ type blob struct {
 	fetchedRegionSet   regionSet
 	fetchedRegionSetMu sync.Mutex
 
+	// regions which are currently being fetched
+	beingFetchedRegions     map[region]struct{}
+	beingFetchedRegionsMu   sync.Mutex
+	beingFetchedRegionsCond sync.Cond
```
ktock, line 73: Sorry for the late review. Could we use singleflight? https://pkg.go.dev/golang.org/x/sync/singleflight

vkuzniet: The change is ready. Still need to benchmark a bunch of images to make sure that there's no regression. Will release from Draft within the next few days.

vkuzniet: All set.

fs/remote/blob.go
```diff
@@ -281,89 +310,99 @@ func (b *blob) fetchRange(allData map[region]io.Writer, opts *options) error {
...
-	mr, err := fr.fetch(ctx, req, true, opts)
-	if err != nil {
-		return err
-	}
-	defer mr.Close()
+	// We build a key based on regions we need to fetch and pass it to singleflightGroup.Do(...)
+	// to block simultaneous same requests. once the request is finished and the data is ready,
+	// all blocked callers will be unblocked and that same data will be returned by all blocked callers.
+	key := makeSyncKey(allData)
+	_, err, _ := b.fetchedRegionGroup.Do(key, func() (interface{}, error) {
```
ktock, line 317: Writes to `allData map[region]io.Writer` need to occur even for blocked callers otherwise `ReadAt` will return invalid contents. 
Blocked callers should try to acquire contents from the cache after resuming. If they can't get the contents from the cache, they make requests to the registry.

fs/remote/blob_test.go
```diff
@@ -278,40 +280,213 @@ func TestFailReadAt(t *testing.T) {
...
+		)
+
+		start := make(chan struct{})
+		wg.Add(routines)
+
+		for i := 0; i < routines; i++ {
+			allData := make(map[region]io.Writer)
+			if i < len(tst.regions) {
+				for j := range tst.regions[i] {
+					allData[tst.regions[i][j]] = ioutil.Discard
```
ktock, line 449: Could we add the check for the contents as well?

fs/remote/blob.go
vkuzniet: Done.

fs/remote/blob_test.go
vkuzniet: Done.

ktock (commented):
Thank you. Overall looks good. 

Some minor nits:

fs/remote/blob.go
```diff
@@ -64,15 +67,33 @@ type blob struct {
...
+		chunkSize:         chunkSize,
+		prefetchChunkSize: prefetchChunkSize,
+		cache:             blobCache,
+		lastCheck:         lastCheck,
+		checkInterval:     checkInterval,
+		resolver:          r,
+		fetchTimeout:      fetchTimeout,
+	}
+}
+
```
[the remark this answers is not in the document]
vkuzniet: That function was born because of complexity to create `blob` object during previous operations. Even though the object creation complexity is back to normal, I would prefer to keep the function, since this allows us to extend `blob` without worrying about implementing `makeBlob` again.

fs/remote/blob_test.go
```diff
@@ -278,40 +280,245 @@ func TestFailReadAt(t *testing.T) {
...
 	}
 }
 
-func makeBlob(t *testing.T, size int64, chunkSize int64, prefetchChunkSize int64, fn RoundTripFunc) *blob {
-	return &blob{
-		fetcher: &fetcher{
+func TestParallelDownloadingBehavior(t *testing.T) {
+	type testData struct {
+		name           string
+		regions        [][]region
```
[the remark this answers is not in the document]
vkuzniet: Done.

fs/remote/blob.go
```diff
@@ -153,6 +174,15 @@ func (b *blob) FetchedSize() int64 {
 	return sz
 }
 
+func makeSyncKey(allData map[region]io.Writer) string {
+	var keys []string
+	for key := range allData {
+		keys = append(keys, fmt.Sprintf("[%d,%d]", key.b, key.e))
+	}
```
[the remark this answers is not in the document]
vkuzniet: Thanks for the tip. All set.

fs/remote/blob_test.go
```diff
@@ -278,40 +280,245 @@ func TestFailReadAt(t *testing.T) {
...
+				},
+				chunkSize: tst.chunkSize,
+				size:      int64(len(tst.content)),
+				cache:     cache.NewMemoryCache(),
+			}
+		)
+
+		start := make(chan struct{})
+		wg.Add(routines)
+		p := make([]byte, len(tst.content))
```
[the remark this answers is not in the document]
vkuzniet: Yes, they were. Updated the test to avoid that.

ktock (approved):
LGTM. Thank you!

PR #448, opened 2021-09-03
Reset bytes.Buffer before returning it to sync.Pool

kzys:
Most of bytes.Buffer in stargz-snapshotter is unbounded in terms of
size and sync.Pool's Get may choose to ignore the pool.

Resizing a buffer before returning to the pool may alleviate the
snapshotter's memory utilization issue.

https://github.com/golang/go/issues/23199 has a long discussion about
the issue.

### DIFF
--- a/cache/cache.go
+++ b/cache/cache.go
@@ -132,6 +132,7 @@ func NewDirectoryCache(directory string, config DirectoryCacheConfig) (BlobCache
 		}
 		dataCache = lrucache.New(maxEntry)
 		dataCache.OnEvicted = func(key string, value interface{}) {
+			value.(*bytes.Buffer).Reset()
 			bufPool.Put(value)
 		}
 	}
@@ -296,7 +297,6 @@ func (dc *directoryCache) Add(key string, opts ...Option) (Writer, error) {
 	}
 
 	b := dc.bufPool.Get().(*bytes.Buffer)
-	b.Reset()
 	memW := &writer{
 		WriteCloser: nopWriteCloser(io.Writer(b)),
 		commitFunc: func() error {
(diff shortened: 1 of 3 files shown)


cache/cache.go
```diff
@@ -331,14 +331,19 @@ func (dc *directoryCache) Add(key string, opts ...Option) (Writer, error) {
...
+			dc.putBuffer(b) // abort it.
 			return nil
 		},
 	}
 
 	return memW, nil
 }
 
+func (dc *directoryCache) putBuffer(b *bytes.Buffer) {
+	b.Reset()
```
kzys, line 343: Note that calling Reset() is the safest but may not be the most efficient solution.

Go's standard library has some more sophisticated examples.

https://github.com/golang/go/blob/9e7bc80b31088dc62faf4776ffdb1a2e27afa94e/src/os/dir_unix.go#L30-L36

https://github.com/golang/go/blob/acc2957bc9873ab7e65942045830a3dc0592eda2/src/net/http/server.go#L794-L850

fs/reader/reader.go
```diff
@@ -406,11 +411,10 @@ func (sf *file) ReadAt(p []byte, offset int64) (int, error) {
 
 		// Use temporally buffer for aligning this chunk
 		b := sf.gr.bufPool.Get().(*bytes.Buffer)
-		b.Reset()
```
ktock, line 409: Could we remain `Reset` before use for safety?

kzys: Sure, but would it have real effects? Get() returns either a new Buffer, which is safe to use according to https://cs.opensource.google/go/go/+/refs/tags/go1.17:src/bytes/buffer.go;l=449

> In most cases, new(Buffer) (or just declaring a Buffer variable) is sufficient to initialize a Buffer.

 or a buffer which is put by putBuffer().

ktock: Sorry, I think you are right. Merging.

ktock (approved):

PR #453, opened 2021-09-09
estargz: support lossless compression

ktock:
https://github.com/moby/buildkit/pull/2352

Currently, eStargz writer (`estargz.Writer`) rewrites tar headers of the original tar so it's not possible to recover the original tar blob from the result eStargz. This lossy behaviour can be problematic when the client needs to convert tar to eStargz and then recover the original tar from the result eStargz.

For solving this issue, this commit adds `AppendTarLossLess` API to `estargz.Writer`. This allows creating eStargz without changing the original tar blob. `estargz.Unpack` API enables to decompress the eStargz and get the original tar blob.

Note that this losslessness implies a limitation that the writer cannot support compressing a tar that contains a file named `stargz.index.json` because estargz cannot be created without overwriting that file. When the caller appends an eStargz blob to the writer, it should be decompressed to a tar using `estargz.Unpack` API.

cc @tonistiigi 
Thank you for the discussion.


estargz/testutil.go
```diff
@@ -148,93 +148,96 @@ func testBuild(t *testing.T, controllers ...TestingController) {
...
-						)
-						if err != nil {
-							t.Fatalf("failed to parse the want stargz: %v", err)
-						}
+							// Prepare sample data
+							wantBuf := new(bytes.Buffer)
+							sw := NewWriterWithCompressor(wantBuf, cl)
+							sw.ChunkSize = tt.chunkSize
+							if err := sw.AppendTar(tarBlob); err != nil {
+								t.Fatalf("faield to append tar to want stargz: %v", err)
```
AkihiroSuda, line 182: nit: typo

AkihiroSuda (approved):

PR #479, opened 2021-09-28
Make HTTP requests retryable

rdpsin:
Currently, the snapshotter does not support retries when sending out HTTP requests.
Any kind of intermittent networking error (such as networking failures or error status codes)
can potentially disrupt existing container workloads.

This patch uses [go-retryablehttp](https://github.com/hashicorp/go-retryablehttp) to implement retries in case of network error.
It also handles cases where the server replies with either a 429 or 5xx status code.

### DIFF
--- a/fs/config/config.go
+++ b/fs/config/config.go
@@ -69,6 +69,10 @@ type BlobConfig struct {
 	// If PrefetchChunkSize < ChunkSize prefetch bytes will be fetched as a single http GET,
 	// else total GET requests for prefetch = ceil(PrefetchSize / PrefetchChunkSize).
 	PrefetchChunkSize int64 `toml:"prefetch_chunk_size"`
+
+	MaxRetries  int `toml:"max_retries"`
+	MinWaitMSec int `toml:"min_wait_msec"`
+	MaxWaitMSec int `toml:"max_wait_msec"`
 }
 
 type DirectoryCacheConfig struct {
--- a/fs/remote/resolver.go
+++ b/fs/remote/resolver.go
@@ -28,6 +28,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"math/rand"
 	"mime"
 	"mime/multipart"
 	"net/http"
@@ -46,15 +47,21 @@ import (
 	"github.com/containerd/stargz-snapshotter/fs/config"
 	commonmetrics "github.com/containerd/stargz-snapshotter/fs/metrics/common"
 	"github.com/containerd/stargz-snapshotter/fs/source"
+	rhttp "github.com/hashicorp/go-retryablehttp"
 	digest "github.com/opencontainers/go-digest"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 )
 
 const (
 	defaultChunkSize        = 50000
 	defaultValidIntervalSec = 60
 	defaultFetchTimeoutSec  = 300
+
+	defaultMaxRetries  = 5
+	defaultMinWaitMSec = 30
+	defaultMaxWaitMSec = 300000
 )
 
 func NewResolver(cfg config.BlobConfig) *Resolver {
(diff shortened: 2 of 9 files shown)


service/resolver/cri.go
```diff
@@ -136,18 +137,20 @@ func RegistryHostsFromCRIConfig(ctx context.Context, config Registry, credsFuncs
 			}
 
 			var (
-				transport = http.DefaultTransport.(*http.Transport).Clone()
-				client    = &http.Client{Transport: transport}
-				config    = config.Configs[u.Host]
+				rclient = rhttp.NewClient()
```
ktock, line 140: Why not `client = rhttp.NewClient().StandardClient()` here?

fs/config/config.go
```diff
@@ -69,6 +69,10 @@ type BlobConfig struct {
 	// If PrefetchChunkSize < ChunkSize prefetch bytes will be fetched as a single http GET,
 	// else total GET requests for prefetch = ceil(PrefetchSize / PrefetchChunkSize).
 	PrefetchChunkSize int64 `toml:"prefetch_chunk_size"`
+
+	MaxRetries  int `toml:"max_retries"`
+	MinWaitTime int `toml:"min_wait_time"` // In milliseconds.
+	MaxWaitTime int `toml:"max_wait_time"` // In milliseconds.
```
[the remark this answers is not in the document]
rdpsin: Sure, I'll update it.

service/resolver/cri.go
rdpsin: So that we can set `TLSClientConfig` on the `http.Transport` that is a field of `HTTPClient` (which is a field of `retryablehttp.Client`) using `getTLSConfig`.

Equivalent of doing:

```
client := rhttp.NewClient().StandardClient()
client.Transport.(*rhttp.RoundTripper).Client.HTTPClient.Transport.(*http.Transport).TLSConfig = getTLSConfig(*config.TLS)
```

service/resolver/cri.go
```diff
@@ -136,18 +137,20 @@ func RegistryHostsFromCRIConfig(ctx context.Context, config Registry, credsFuncs
...
+				config  = config.Configs[u.Host]
 			)
 
 			if config.TLS != nil {
-				transport.TLSClientConfig, err = getTLSConfig(*config.TLS)
-				if err != nil {
-					return nil, errors.Wrapf(err, "get TLSConfig for registry %q", e)
+				if tr, ok := rclient.HTTPClient.Transport.(*http.Transport); ok {
```
ktock, line 145: Should return an error if the tls config cannot be applied (i.e. `!ok`).

rdpsin: Thanks, I'll update it.

rdpsin: Fixed.

fs/config/config.go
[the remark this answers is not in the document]
rdpsin: Changed to `MinWaitMSec` and `MaxWaitMSec`

ktock:
Could you apply `go mod tidy`?

https://github.com/containerd/stargz-snapshotter/pull/479/checks?check_run_id=3739646455

```
  These files were modified:
  
   M go.sum
```

fs/remote/resolver.go
```diff
@@ -120,13 +182,23 @@ func newFetcher(ctx context.Context, hosts source.RegistryHosts, refspec referen
...
 		// Prepare transport with authorization functionality
 		tr := host.Client.Transport
+
+		if rt, ok := tr.(*rhttp.RoundTripper); ok {
+			rt.Client.RetryMax = fc.maxRetries
+			rt.Client.RetryWaitMin = fc.minWaitMSec
+			rt.Client.RetryWaitMax = fc.maxWaitMSec
+			rt.Client.Backoff = backoffStrategy
+			rt.Client.CheckRetry = retryStrategy
+			rt.Client.Logger = nil // disable logging every request
```
ktock, line 199: Logger should be disabled when the RoundTripper is created (i.e. `rhttp.NewClient()`).

rdpsin: Done. Thanks.

ktock (approved):
Thank you!

PR #523, opened 2021-11-18
Bump up containerd to the latest

ktock:
See https://github.com/containerd/containerd/commit/a776a27af54a803657d002e7574a4425b3949f56

Also fixes codes relaeted to mediatype detection to make it stricter.

### DIFF
--- a/util/containerdutil/manifest.go
+++ b/util/containerdutil/manifest.go
@@ -19,6 +19,7 @@ package containerdutil
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"sort"
 
 	"github.com/containerd/containerd/content"
@@ -40,6 +41,9 @@ func ManifestDesc(ctx context.Context, provider content.Provider, image ocispec.
 		case images.MediaTypeDockerSchema2Manifest, ocispec.MediaTypeImageManifest:
 			p, err := content.ReadBlob(ctx, provider, desc)
 			if err != nil {
+				return nil, err
+			}
+			if err := ValidateMediaType(p, desc.MediaType); err != nil {
 				return nil, err
 			}
 			var manifest ocispec.Manifest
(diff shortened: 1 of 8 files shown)


util/containerdutil/manifest.go
```diff
@@ -109,3 +116,43 @@ func ManifestDesc(ctx context.Context, provider content.Provider, image ocispec.
 	}
 	return m[0], nil
 }
+
+// Forked from github.com/containerd/containerd/images/image.go
```
AkihiroSuda, line 120: specify the hash or tag

util/containerdutil/manifest.go
```diff
@@ -109,3 +116,43 @@ func ManifestDesc(ctx context.Context, provider content.Provider, image ocispec.
...
+	Config    json.RawMessage `json:"config,omitempty"`
+	Layers    json.RawMessage `json:"layers,omitempty"`
+	Manifests json.RawMessage `json:"manifests,omitempty"`
+	FSLayers  json.RawMessage `json:"fsLayers,omitempty"` // schema 1
+}
+
+// ValidateMediaType returns an error if the byte slice is invalid JSON or if
+// the media type identifies the blob as one format but it contains elements of
+// another format.
+func ValidateMediaType(b []byte, mt string) error {
```
AkihiroSuda, line 135: Can we just update containerd to expose https://github.com/containerd/containerd/blob/a776a27af54a803657d002e7574a4425b3949f56/images/image.go#L401 as a public function?

Can be a follow-up PR.

util/containerdutil/manifest.go
samuelkarp, line 120: Is it worth exposing the original implementation rather than copying?  This logic covers the known cases of OCI indices, OCI manifests, Docker manifest lists, Docker schema 2 manifests, and Docker schema 1 manifests, but may have to change in the future if new types are introduced.

AkihiroSuda: Opened PR https://github.com/containerd/containerd/pull/6266

util/containerdutil/manifest.go
AkihiroSuda: Opened PR https://github.com/containerd/containerd/pull/6266

samuelkarp: > Can be a follow-up PR.

:+1: to this, though it'd be worth opening an issue to track it.

AkihiroSuda (approved):

samuelkarp (approved):

Issue #509, opened 2021-11-02
Add test coverage report for Pull Request


gtxu:
For now we have a unit test for all component before PR get merged, but we don't have unit test coverage report for a single PR. There are tools (such as [Coveralls](https://coveralls.io/))that automatically post unit test coverage report for a PR in comment. This may be a considerable option for stargz-snapshotter.

gtxu:
An example of coveralls
<img width="1203" alt="Screen Shot 2021-11-02 at 3 12 43 PM" src="https://user-images.githubusercontent.com/91280642/139931410-ab6117e6-d2fa-4e63-aada-558dc91ae5fc.png">
:

vkuzniet:
Hi ktock@, what do you think about adding this to the existing pipeline of stargz snapshotter?

PR #536, opened 2021-11-22
add unit test coverage in GitHub workflow

gtxu:
This plug-in added in Github workflow will trigger a comment from coveralls.io when each PR, showing the unit test coverage information and impact on test coverage that PR made.
This will resolve [Issue 509](https://github.com/containerd/stargz-snapshotter/issues/509)

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -21,6 +21,14 @@ jobs:
     - name: Test all
       run: ./script/util/make.sh test-all -j2
 
+  test-coverage:
+    runs-on: ubuntu-20.04
+    name: Coveralls
+    steps:
+    - uses: coverallsapp/github-action@master
+      with:
+        github-token: ${{ secrets.GITHUB_TOKEN }}
+
   linter:
     runs-on: ubuntu-20.04
     name: Linter


.github/workflows/tests.yml
```diff
@@ -21,6 +21,14 @@ jobs:
     - name: Test all
       run: ./script/util/make.sh test-all -j2
 
+  test-coverage:
+    runs-on: ubuntu-20.04
+    name: Coveralls
+    steps:
+    - uses: coverallsapp/github-action@master
+      with:
+        github-token: ${{ secrets.GITHUB_TOKEN }}
```
ktock, line 30: I don't think we need this in our CI. Users can check the test coverage using `go test -cover`.

cc @AkihiroSuda

ktock: This thread might be helpful:  https://stackoverflow.com/questions/10516662/how-to-measure-test-coverage-in-go

ktock: But thank you very much for your suggestion! :smile:

gtxu: @ktock Thank you for your feedback!

PR #548, opened 2021-12-03
feat: log success record path when debug

jonyhy96:
Currently, it will print path only when error happens. Add log when success record path could help to see which file will be prefetched

### DIFF
--- a/analyzer/analyzer.go
+++ b/analyzer/analyzer.go
@@ -200,6 +200,10 @@ func Analyze(ctx context.Context, client *containerd.Client, ref string, opts ..
 	var fanotifierClosed bool
 	var fanotifierClosedMu sync.Mutex
 	go func() {
+		var successCount int
+		defer func() {
+			log.G(ctx).Debugf("success record %d path", successCount)
+		}()
 		for {
 			path, err := fanotifier.GetPath()
 			if err != nil {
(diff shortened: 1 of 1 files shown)


jonyhy96:
PTAL @ktock

analyzer/analyzer.go
```diff
@@ -217,6 +217,7 @@ func Analyze(ctx context.Context, client *containerd.Client, ref string, opts ..
 			if err := rc.Record(path); err != nil {
 				log.G(ctx).WithError(err).Debugf("failed to record %q", path)
 			}
+			log.G(ctx).Debugf("success record path %q", path)
```
ktock, line 220: You can get the full list of recorded path using `--record-out` option.

Instead of printing all paths, can we just print the `succeeded` message with the number of recorded path just before this `go func` returns?

jonyhy96: Done, do u think we should add `use with --record-out option to see details` with the log?

analyzer/analyzer.go
```diff
@@ -200,6 +200,10 @@ func Analyze(ctx context.Context, client *containerd.Client, ref string, opts ..
 	var fanotifierClosed bool
 	var fanotifierClosedMu sync.Mutex
 	go func() {
+		var successCount int
+		defer func() {
+			log.G(ctx).Debugf("success record %d path", successCount)
```
ktock, line 205:
```suggestion
			log.G(ctx).Debugf("recorded %d path", successCount)
```

ktock (approved):

ktock:
Thanks

PR #600, opened 2022-01-21
layer resolver: Avoid many cache misses occur when many pullings of images happen

ktock:
When `filesystem.Mount` is called for the first layer in an image, stargz-snapshotter resolves all layers in that image in parallel and caches these resolved layers metadata in LRU cache. When `filesystem.Mount` is called for the neighbouring layers of that image, the cached layer can be used for speeding up the mounts.

However, when many pulling of images happen in parallel, layers are soonly evicted from the LRU cache by other image pullings and many cache misses happen. This result to many resource consumption (e.g. fd), many (duplicated) requests to the registry, etc.

This commit solves this by using TTL-based cache instead of LRU cache. TTL cache doesn't have size limitations and manages eviction using TTL of each element. This avoids the above problems of quick evictions and many cache misses when many parallel pulling of images happen. When `filesystem.Mount` of a layer of an image is called, mounts of the neighbouring layers also happen soon. And once a mount of a layer completes, reusing of the layer is managed by the snapshotter's side but not by the filesystem so we don't need to cache the layer in long term. So TTL-based cache should be a better choice than LRU cache here.

## max fd consumption comparison

The following client command mounts 7 images (74 snapshots) in parallel.

```
TARGETS="gcc:10.2.0 golang:1.12.9 jenkins:2.60.3 node:13.13.0 php:7.3.8 python:3.9 tomcat:10.0.0-jdk15-openjdk-buster"
for I in $TARGETS ; do
  ctr-remote i rpull --plain-http registry2:5000/${I}-esgz &
done
```

|branch|max fd consumption|
|---|---|
|main|381|
|This PR|211|

Each snapshot consumes 1 fd for fuse, 2 fds (at most) for registry connection.
So stargz-snapshotter consumes around `3 * number_of_snapshots` fds (+ misc fds for local cache, etc.).
Main branch consumes much more fds than expected because of lots of cache misses and duplicated layer resolves.

### DIFF
--- a/util/cacheutil/ttlcache_test.go
+++ b/util/cacheutil/ttlcache_test.go
@@ -1,0 +1,178 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package cacheutil
+
+import (
+	"sync"
+	"testing"
+	"time"
+)
+
+// TestTTLAdd tests Add API
+func TestTTLAdd(t *testing.T) {
+	c := NewTTLCache(time.Hour)
+
+	key, value := "key1", "abcd"
+	v, _, added := c.Add(key, value)
+	if !added {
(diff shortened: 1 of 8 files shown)


util/cacheutil/ttlcache_test.go
```diff
@@ -0,0 +1,196 @@
...
+	time.Sleep(3 * time.Second) // wait until elements reach ttl
+
+	evictedMu.Lock()
+	if len(evicted) != 1 {
+		t.Errorf("1 content must be removed")
+		return
+	}
+	if evicted[0] != key1 {
+		t.Errorf("1st content %q must be evicted but got %q", key1, evicted[0])
+		return
```
AkihiroSuda, line 171: Can we just use t.Fatalf instead of t.Errorf + return

AkihiroSuda (approved):

PR #701, opened 2022-03-23
Migrate from `ioutil` to `io` and `os`

ktock:
`ioutil` has been deprecated in go 1.16.


AkihiroSuda (approved):
LGTM, but we should consider having linter to block ioutil (in another PR)

PR #719, opened 2022-03-30
cmd/ctr/convert: should return error if without --oci

fuweid:
Signed-off-by: Wei Fu <fuweid89@gmail.com>

### DIFF
--- a/cmd/ctr-remote/commands/convert.go
+++ b/cmd/ctr-remote/commands/convert.go
@@ -32,7 +32,6 @@ import (
 	zstdchunkedconvert "github.com/containerd/stargz-snapshotter/nativeconverter/zstdchunked"
 	"github.com/containerd/stargz-snapshotter/recorder"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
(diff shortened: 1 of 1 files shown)


cmd/ctr-remote/commands/convert.go
```diff
@@ -131,7 +130,7 @@ When '--all-platforms' is given all images in a manifest list must be available.
 			}
 			layerConvertFunc = estargzconvert.LayerConvertFunc(esgzOpts...)
 			if !context.Bool("oci") {
-				logrus.Warn("option --estargz should be used in conjunction with --oci")
+				return errors.New("option --estargz should be used in conjunction with --oci")
```
AkihiroSuda, line 133: Not a hard error

fuweid: hmm. But the snapshotter needs that annotations 😂. without the annotations, rpull will fetch all the contents

fuweid: thanks! Since it is not issue, I am close it.

PR #860, opened 2022-07-21
Rename estargz/testutil.go so that it is not included in non-test callers

mtrmac:


mtrmac:
Oops, I have missed that `CompressionTestSuite` is called from outside the package … and from inside … and it needs access to internals. I’m afraid that refactoring that into multiple packages that avoids reference cycles is way more work than I expected.

PR #1011, opened 2022-11-29
Support CRI v1 API

ktock:
This commit adds support for CRI v1 API for CRI proxy. We still support CRI v1alpha API because containerd supports it. We'll drop that API once containerd drops that API.


ktock:
https://github.com/containerd/stargz-snapshotter/actions/runs/3572035809/jobs/6004477580#step:7:1443

```
github.com/k3s-io/k3s/pkg/containerd imports
	github.com/containerd/stargz-snapshotter/service/plugin imports
	github.com/containerd/containerd/third_party/k8s.io/cri-api/pkg/apis/runtime/v1alpha2: module github.com/containerd/containerd@latest found (v1.6.10, replaced by github.com/k3s-io/containerd@v1.5.14-k3s1), but does not contain package 
```

It seems that we need to maintain or own CRI API fork.

ktock:
Added `github.com/containerd/stargz-snapshotter/third_party/k8s.io/cri-api/pkg/apis/runtime/v1alpha2` as a fork of  CRI API.

ktock:
Here we introduce `github.com/containerd/stargz-snapshotter/service/pluginforked`.

Both of `github.com/containerd/stargz-snapshotter/service/pluginforked` and the current `github.com/containerd/stargz-snapshotter/service/plugin` register our stargz snapshotter plugin to the containerd plugin system (e.g. used by k3s). So most of user can use  `github.com/containerd/stargz-snapshotter/service/plugin` and ignore `github.com/containerd/stargz-snapshotter/service/pluginforked`.

`github.com/containerd/stargz-snapshotter/service/plugin` depends on containerd newer than [234bf990dca4e81e89f549448aa6b555286eaa7a](https://github.com/containerd/containerd/commit/234bf990dca4e81e89f549448aa6b555286eaa7a) which contains CRI v1alpha API fork. So if the user imports stargz-snapshotter to the tool that imports older version of containerd, the user can't build it. In this case, the user needs to use `github.com/containerd/stargz-snapshotter/service/pluginforked` instead. This plugin contains our forked CRI v1alpha API.

The latest k3s (`953c7699b413f421a9543d45d282a1d3ce8e3598`) doesn't imports the recent version of containerd so they need to use `github.com/containerd/stargz-snapshotter/service/pluginforked` until upgrading containerd. Our k3s integration test uses `github.com/containerd/stargz-snapshotter/service/pluginforked` as well.

Once the user upgrade containerd to newer than 234bf990dca4e81e89f549448aa6b555286eaa7a, the uesr can't use `github.com/containerd/stargz-snapshotter/service/pluginforked` and protobuf code fails with the error e.g. `proto: duplicate enum registered: runtime.v1alpha2.Protocol`. In this case,`github.com/containerd/stargz-snapshotter/service/plugin` need to be used.

AkihiroSuda (approved):

PR #1067, opened 2023-01-20
ipfs: Use net/http instead of binary to access IPFS API

ktock:
By this commit, we don't need ipfs binary for lazy pulling (ipfs daemon still needs running). Similar fix will be applied to nerdctl later.


ipfs/converter.go
```diff
@@ -104,3 +108,29 @@ func GetCID(desc ocispec.Descriptor) (string, error) {
...
+	}
+	a, err := ma.NewMultiaddr(strings.TrimSpace(string(api)))
+	if err != nil {
+		return "", err
+	}
+	_, url, err := manet.DialArgs(a)
+	if err != nil {
+		return "", err
+	}
+	return "http://" + url, nil // TODO: support other protocols?
```
AkihiroSuda, line 135: No support for https?

AkihiroSuda: Is this a regression, or did we never support https?

ipfs/client.go
```diff
@@ -0,0 +1,167 @@
+/*
```
AkihiroSuda, line 1: Maybe split the package to `ipfs/client/client.go`, so that it can be imported by third party projects without dependency on containerd pkgs

ipfs/client.go
```diff
@@ -0,0 +1,167 @@
...
+	client := http.DefaultClient
+	resp, err := client.Do(req)
+	if err != nil {
+		return "", err
+	}
+	defer func() {
+		io.Copy(io.Discard, resp.Body)
+		resp.Body.Close()
+	}()
+	if resp.StatusCode != 200 {
```
AkihiroSuda, line 154: Some server may potentially return other 2xx codes?

ipfs/client.go
```diff
@@ -0,0 +1,167 @@
...
+	var rs FileInfo
+	if err := json.NewDecoder(resp.Body).Decode(&rs); err != nil {
+		return nil, err
+	}
+	return &rs, nil
+}
+
+// Cat get the reader of the data specified by the IPFS path and optionally with
+// the offset and length.
+func Cat(ipfsAPI string, p string, offset *int, length *int) (_ io.ReadCloser, retErr error) {
```
AkihiroSuda, line 76: Why cat? This doesn't seem to support conCATination

ipfs/client.go
ktock: Moved the package to ipfs/client/client.go

ipfs/client.go
ktock: Fixed to allow 2xx.

ipfs/client.go
ktock: Renamed to "Get".

ipfs/converter.go
ktock: Fixed `ipfs/client/client.go` to allow other protocols. Stargz snapshotter still doesn't support https but I think this isn't a regression because when we used [`NewLocalApi` of `go-ipfs-http-client`](https://github.com/ipfs/go-ipfs-http-client/blob/171fcd55e3b743c38fb9d78a34a3a703ee0b5e89/api.go#L45), it didn't seem to support https as well (it seems that go-ipfs won't support https?: https://discuss.ipfs.tech/t/installing-ssl-certificate/388/2).

ipfs/client/client.go
```diff
@@ -0,0 +1,228 @@
...
+type Client struct {
+	// Address is URL of IPFS API to connect to.
+	Address string
+
+	// Client is http client to use for connecting to IPFS API
+	Client *http.Client
+}
+
+// NewClient creates a new IPFS API client of the specified address.
+func NewClient(ipfsAPIAddress string) *Client {
```
AkihiroSuda, line 45:
```suggestion
func New(ipfsAPIAddress string) *Client {
```

script/ipfs/entrypoint.sh
```diff
@@ -0,0 +1,45 @@
...
+    if [ "${SUCCESS}" == "true" ] ; then
+        return 0
+    else
+        return 1
+    fi
+}
+
+ipfs init
+ipfs daemon --offline &
+retry curl -X POST localhost:5001/api/v0/version >/dev/null 2>&1 # wait for up
```
AkihiroSuda, line 43: Probably you can just use `curl --retry N --retry-all-errors --retry-delay SECONDS`

AkihiroSuda (approved):

Issue #1180, opened 2023-03-27
`remote.Resolver` timeout doesn't work as expected


Kern--:
While trying to write an integration test for poor network conditions in the [soci-snapshotter](https://github.com/awslabs/soci-snapshotter), we noticed that the timeout logic didn't work the way we expected.

In particular, we found that when a connection timed out, no retries were performed and the request was abandonded after the first failure. We also noticed that no matter what we set the default timeout to, the request would always block for 30 seconds. We expected that a connection timeout would retry like any other http failure and that the default timeout would be the timeout for each request.

### Root cause

The root cause appears to be the way that retries were added to the stargz-snapshotter:
https://github.com/containerd/stargz-snapshotter/blob/0c9f87672e4e93e9a0c8819a0911854406b3217c/service/resolver/registry.go#L67-L76

The structure of `tr` after this block is the following (ignoring irrelevant fields):
```
http.Client {
    Timeout: <configured timeout>
    Transport: rttp.Transport {
        Client: rhttp.Client {
            HTTPClient: http.Client {}
        }
    }
}
```

What appears to be happening is that once the outer `http.Client` timeout is reached, the request context is cancelled which prevents the `rhttp.Client` from doing any retries. 

What also seems to be happening is that the inner `http.Client` is not selecting over the request context at all because it doesn't have a timeout. The 30s before it exits appears to be a dial timeout configured deep in the default client used by go-retryablehttp https://github.com/hashicorp/go-cleanhttp/blob/master/cleanhttp.go#L30. 

I think the solution here is to move the timeout to the inner http client so that it defines the per-request timeout rather than the total request timeout. The total request timeout is then configurable based on the retry policy. 

### Related Issues

There are a couple of related issues that might be worth tracking too:
1) The default timeout is hardcoded to 30s https://github.com/containerd/stargz-snapshotter/blob/main/service/resolver/registry.go#L30
2) Registry mirrors have a configurable timeout https://github.com/containerd/stargz-snapshotter/blob/main/service/resolver/registry.go#L52, but the main host itself doesn't https://github.com/containerd/stargz-snapshotter/blob/main/service/resolver/registry.go#L65
[...]

PR #1181, opened 2023-03-27
Make timeout per-request

Kern--:
Part of https://github.com/containerd/stargz-snapshotter/issues/1180

When setting up an rhttp client, the structure looks like this:
```
http.Client {
    Transport: rhttp.Transport {
        Client: rhttp.Client {
            HTTPClient: http.Client{}
        }
    }
}
```

Before this change, the timeout was set on the outer client. When the timeout was reached, the request context was cancelled and no more retries would be attempted by the retryable client. Effectively the timeout was being set for the entire retryable request.

This change moves the timeout to the inner client, making the timeout per-request and the timeout of the entire retryable request is governed by the retry policy.

### DIFF
--- a/fs/remote/resolver.go
+++ b/fs/remote/resolver.go
@@ -215,15 +215,16 @@ func newHTTPFetcher(ctx context.Context, fc *fetcherConfig) (*httpFetcher, int64
 		// Prepare transport with authorization functionality
 		tr := host.Client.Transport
 
+		timeout := host.Client.Timeout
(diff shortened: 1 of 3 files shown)


ktock:
@Kern-- Thank you for the patch, looks good to me. Could you rebase thie patch on the latest main branch for fixing the CI failure?

ktock (approved):

ktock:
Thanks!

PR #1326, opened 2023-07-26
fix: rollback snapshot to prevent bolt deadlock

goller:
The defer function above checks for the err in the outer scope to determine if it should rollback.
The error in the inner-scope of isRemote is a shadow and when it is an error will not cause a rollback of the transaction.

This causes the bolt transaction to keep the write lock open forever.  Any other attempts at taking a write lock will never succeed.

It is not uncommon for `fs.DiskUsage` to be context canceled as it can take a long time to walk the filesystem.

### DIFF
--- a/snapshot/snapshot.go
+++ b/snapshot/snapshot.go
@@ -316,8 +316,9 @@ func (o *snapshotter) commit(ctx context.Context, isRemote bool, name, key strin
 		return err
 	}
 
+	rollback := true
 	defer func() {
-		if err != nil {
+		if rollback {
 			if rerr := t.Rollback(); rerr != nil {
 				log.G(ctx).WithError(rerr).Warn("failed to rollback transaction")
 			}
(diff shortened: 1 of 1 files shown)


snapshot/snapshot.go
```diff
@@ -331,7 +331,8 @@ func (o *snapshotter) commit(ctx context.Context, isRemote bool, name, key strin
 	}
 
 	if !isRemote { // skip diskusage for remote snapshots for allowing lazy preparation of nodes
-		du, err := fs.DiskUsage(ctx, o.upperPath(id))
+		var du fs.Usage
+		du, err = fs.DiskUsage(ctx, o.upperPath(id))
```
ktock, line 335: Could you add a comment line of notifying that we need to expose `err` to the outer scope?

ktock (commented):
Thanks! Could you sign your commit with `Signed-off-by`?

snapshot/snapshot.go
goller: @ktock Will do !

goller: @ktock I went a bit further and used a pattern you used in the `createSnapshot` function.  I like what you did there as it has an explicit variable rather than having an err with double duty.

https://github.com/containerd/stargz-snapshotter/blob/9e9e7a507fc58ab7b01ee93401a3ee2461217a40/snapshot/snapshot.go#L517-L524

PR #1340, opened 2023-08-10
WIP

rbpdt:
Proof of concept for a suggested change to [containerd overlay snapshotter](https://github.com/containerd/containerd/compare/main...pdtpartners:containerd:feature/make-overlay-snapshotter-embeddable).


ktock:
Thanks for the patch. This seems to depend on the on-going patch on containerd (https://github.com/containerd/containerd/pull/8945 ) so let me turn this into a draft PR as of now.

PR #1381, opened 2023-09-12
Protect node.ents and node.entsCached with a mutex in fs/layer/node.go

iain-macdonald:
We run a modified version of the soci-snapshotter (which is based on the stargz-snapshotter) in production at BuildBuddy and recently enabled the golang race detector in our dev environment. It flagged racy accesses to node.ents and node.entsCached, here are the [logs](https://github.com/containerd/stargz-snapshotter/files/12591490/datarace.txt) (line numbers should be the same). This commit protects access to those two fields with a mutex.


ktock (approved):
Thanks!

PR #1387, opened 2023-09-16
Set fuse.Attr.Blocks to number of 512-byte blocks

### LINKED ISSUE #1386: stargz-snapshotter incorrectly reports number blocks for a file.
When reporting the number of blocks in a go-fuse `fuse.Attr`, stargz-snapshotter reports the number of `blockSize`-byte blocks
https://github.com/containerd/stargz-snapshotter/blob/7275d45b185cd720ce13747f0d0a78e9e7cd60e3/fs/layer/node.go#L655-L658

The go-fuse documentation indicates this should be the number of 512-byte blocks

https://github.com/hanwen/go-fuse/blob/043296a854b6094df7937a992ba1426e1bb3c306/fuse/types_linux.go#L22

---

We found this in the soci-snapshotter when a couple users reported their containers hanging after creating SOCI indexes. From strace, we found that the entrypoint scripts were getting stuck in an lseek loop while trying to `cp` a file like:

```
lseek(3, 1437832, SEEK_SET)             = 1437832
lseek(3, 1437832, SEEK_DATA)            = 1437832
lseek(3, 1437832, SEEK_HOLE)            = 1437832
lseek(3, 1437832, SEEK_SET)             = 1437832
lseek(3, 1437832, SEEK_DATA)            = 1437832
lseek(3, 1437832, SEEK_HOLE)            = 1437832
lseek(3, 1437832, SEEK_SET)             = 1437832
```

We tracked this down to a bug in go-fuse where `lseek` with `SEEK_HOLE` at the end of the file should return ENXIO, but go-fuse was returning the current file position. ([Upstream issue](https://github.com/hanwen/go-fuse/issues/487) + [pending PR to fix it](https://github.com/hanwen/go-fuse/pull/488))

While fixing go-fuse is good and solves the problem, the files shouldn't be reporting as sparse so this `lseek` behavior shouldn't be happening at all. We tracked that down to the fact that the `cp` command is detecting the file is sparse because `blocks * 512 < size` which is because SOCI (and stargz) report a lower number of `blockSize`-byte blocks. 

It looks like this lseek behavior changed in `cp` in coreutils 9.0 which is why we aren't seeing this with all containers:

>  cp, install and mv now use the copy_file_range syscall if available.
  Also, they use lseek+SEEK_HOLE rather than ioctl+FS_IOC_FIEMAP on sparse
  files, as lseek is simpler and more portable.

https://savannah.gnu.org/news/?id=10016

---

I reproduced this with stargz building the tip of main and running a ubuntu:23.04 container that I converted. Running this command will hang forever.
```
nerdctl run --snapshotter stargz --rm -it --entrypoint cp --net host ubuntu:23.04-esgz /bin/bash /bin/bash2
```

Kern--:
Before this change, stargz-snapshotter set fuse.Attr.Blocks to the number of `blockSize`-byte blocks. go-fuse and linux expect this to be reported as 512-byte blocks.

Since `blockSize` > 512, the number of blocks was under-reported which made the files look sparse to programs that inspected them (e.g. `cp`).

fixes #1386.

---

I ran a ubuntu:23.04 container to see that this fixed the hanging issue:

Before:
```
# stat /bin/bash
  File: /bin/bash
  Size: 1437832   	Blocks: 352        IO Block: 4096   regular file
# cp /bin/bash /bin/bash2 # hangs forever
```

After:
```
# stat /bin/bash
  File: /bin/bash
  Size: 1437832     Blocks: 2816       IO Block: 4096   regular file
# cp /bin/bash /bin/bash2
# stat /bin/bash2
  File: /bin/bash2
  Size: 1437832     Blocks: 2816       IO Block: 4096   regular file

```

### DIFF
--- a/fs/layer/node.go
+++ b/fs/layer/node.go
@@ -51,12 +51,16 @@ import (
 
 const (
 	blockSize         = 4096
-	whiteoutPrefix    = ".wh."
-	whiteoutOpaqueDir = whiteoutPrefix + whiteoutPrefix + ".opq"
-	opaqueXattrValue  = "y"
-	stateDirName      = ".stargz-snapshotter"
-	statFileMode      = syscall.S_IFREG | 0400 // -r--------
-	stateDirMode      = syscall.S_IFDIR | 0500 // dr-x------
+	physicalBlockSize = 512
+	// physicalBlockRatio is the ratio of blockSize to physicalBlockSize.
+	// It can be used to convert from # blockSize-byte blocks to # physicalBlockSize-byte blocks
+	physicalBlockRatio = blockSize / physicalBlockSize
+	whiteoutPrefix     = ".wh."
+	whiteoutOpaqueDir  = whiteoutPrefix + whiteoutPrefix + ".opq"
+	opaqueXattrValue   = "y"
+	stateDirName       = ".stargz-snapshotter"
+	statFileMode       = syscall.S_IFREG | 0400 // -r--------
+	stateDirMode       = syscall.S_IFDIR | 0500 // dr-x------
 )
 
 type OverlayOpaqueType int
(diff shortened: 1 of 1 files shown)


fs/layer/node.go
```diff
@@ -652,10 +656,7 @@ func entryToAttr(ino uint64, e metadata.Attr, out *fuse.Attr) fusefs.StableAttr
 		out.Size = uint64(len(e.LinkName))
 	}
 	out.Blksize = blockSize
-	out.Blocks = out.Size / uint64(out.Blksize)
-	if out.Size%uint64(out.Blksize) > 0 {
-		out.Blocks++
-	}
+	out.Blocks = (out.Size + uint64(out.Blksize) - 1) / uint64(out.Blksize) * physicalBlockRatio
```
Kern--, line 659: I opted to try to keep the blocks reported aligned to `blockSize`. I'm not sure if anything would care if it were unaligned, but it seemed reasonable to do.

ktock:
@Kern-- Thanks! The patch looks good to me. Could you run `go fmt ./fs/layer/node.go` to make the linter happy?

PR #1448, opened 2023-10-30
CI: temporary disable containerd main branch integration test

ktock:
containerd main branch changed plugin API that will be released in containerd v2 and that makes the plugin API not compatible to containerd v1 plugin APIs. We'll migrate to the new API once containerd v2 is released. As of now, just temporary disable the CI test against containerd main branch until we migrate to v2 API.

This is needed to pass CI in all other PRs.


AkihiroSuda (approved):

AkihiroSuda:
Can we merge?

PR #1599, opened 2024-03-11
Avoid access to Forgotten() before initialized

### LINKED ISSUE #1594: ino reuse in newInodeWithID
At BuildBuddy, we run a modified version of the [soci-snapshotter](https://github.com/awslabs/soci-snapshotter) with the `store/` directory from stargz-snapshotter patched in to support podman as described [here](https://www.buildbuddy.io/blog/image-streaming). We recently encountered a bug caused by podman pulling two images in parallel in very quick succession (<1ms). I tracked it down to apparent ino reuse, hitting [this error](https://github.com/containerd/stargz-snapshotter/blob/6e5d5a0d2769b5fa7e72edfd5df403ad37ebb838/store/fs.go#L244) (note: the error message is incorrect, it should say `refnode` instead of `rootnode`, I can send a fix for this). Upon digging deeper, I discovered that the [logic in idMap](https://github.com/containerd/stargz-snapshotter/blob/6e5d5a0d2769b5fa7e72edfd5df403ad37ebb838/store/fs.go#L637-L647) which attempts to reuse inos sometimes offers recently claimed inos for reuse. This is because the check for whether an inode is active calls [`Inode.Forgotten()`](https://github.com/hanwen/go-fuse/blob/e9e7c22af17af4611b5783a16458647088cc8dec/fs/inode.go#L255) which returns `true` for newly created, unused, non-persistent inodes (because lookupcount is 0 and parents is empty).

Practically, this means that if an refnode Inode has been created but not yet returned to the caller, and a second refnode.Lookup call attempts to mint a new ino, then the code will return the same ino for both of those Inodes (because the first one is `releasable()`), causing the fuse operations for the first Inode to route incorrectly, causing errors in stargz-snapshotter. You can tickle this bug by adding a `time.Sleep()` in `newInodeWithID` between when the Inode is created and when it is returned, and then getting that code to run twice in parallel by e.g. pulling images externally.

I think a reasonable fix is to only attempt to reuse inos after `max` loops all the way around, which should create a pretty big buffer time between when inos are initially allocated and when the snapshotter attempts to reuse them. In most cases I suspect that would prevent ino reuse, but maybe there are some cases where we're loop through all 4B inos. I'm open to other suggestions as well. Let me know what you think.

ktock:
Fixes #1594

Our `Lookup()` implementation returns a newly created `*fusefs.Inode` to go-fuse. Then that `*fusefs.Inode` is configured and managed by go-fuse lib.

Though [`*fusefs.Inode.Forgotten()`](https://github.com/hanwen/go-fuse/blob/e9e7c22af17af4611b5783a16458647088cc8dec/fs/inode.go#L258) allows status check of that object, it shouldn't be called until that `*fusefs.Inode` is [fully configured by go-fuse](https://github.com/hanwen/go-fuse/blob/e9e7c22af17af4611b5783a16458647088cc8dec/fs/bridge.go#L209-L210). Otherwize we hit issue like #1594.

To avoid this race, this commit adds a lock for avoiding calling `*fusefs.Inode.Forgotten()` during configuration on-going in go-fuse.

### DIFF
--- a/store/fs.go
+++ b/store/fs.go
@@ -26,6 +26,7 @@ import (
 	"io"
 	"os/exec"
 	"sync"
+	"sync/atomic"
 	"syscall"
 	"time"
 
@@ -58,17 +59,21 @@ const (
 
 func Mount(ctx context.Context, mountpoint string, layerManager *LayerManager, debug bool) error {
 	timeSec := time.Second
-	rawFS := fusefs.NewNodeFS(&rootnode{
-		fs: &fs{
-			layerManager: layerManager,
-			nodeMap:      new(idMap),
-			layerMap:     new(idMap),
-		},
-	}, &fusefs.Options{
-		AttrTimeout:     &timeSec,
-		EntryTimeout:    &timeSec,
-		NullPermissions: true,
-	})
+	fs := &fs{
+		layerManager: layerManager,
+		nodeMap:      new(idMap),
+		layerMap:     new(idMap),
+	}
+	rawFS := &fsLocker{
+		RawFileSystem: fusefs.NewNodeFS(&rootnode{
+			fs: fs,
+		}, &fusefs.Options{
+			AttrTimeout:     &timeSec,
+			EntryTimeout:    &timeSec,
+			NullPermissions: true,
+		}),
+	}
+	fs.fsLocker = rawFS.wLocker()
 	mountOpts := &fuse.MountOptions{
 		AllowOther: true, // allow users other than root&mounter to access fs
 		FsName:     "stargzstore",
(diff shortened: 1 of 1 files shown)


iain-macdonald:
My understanding of the interactions between the go-fuse and stargz-snapshotter parts of the code may be off here, but won't the `RLock`/`RUnlock` calls in fsLocker surrouned the `Lock`/`Unlock` calls nested in `newNodeWithID`, causing deadlock? And if not, then isn't the lock in `releasable` insufficient as it doesn't guard the entire period from creation to initialization?

I think a simpler fix would be to used `Inode.changeCounter` to detect when an Inode hasn't been initialized and pair that with the Forgotten() check that's already present. I asked about this possibility in https://github.com/hanwen/go-fuse/issues/504, so what would you think about seeing what Han-Wen says and then coming up with a more elaborate solution if that doesn't work?

ktock:
Fixed in  #1808.

PR #1705, opened 2024-06-11
store: use Forgotten() for lifetime management with leveraging Initialized() for readiness check

ktock:
Fixes: #1599 
Fixes: #1594

Thanks to @iain-macdonald 's patch for go-fuse (https://github.com/hanwen/go-fuse/pull/520), we can rely on Inode.Forgotten to check if that inode is reusable or not. This commit cherry-picks @iain-macdonald's patch to https://github.com/ktock/go-fuse/commit/f59a74eb644ad60f419be455498a183bcd1e1f4f and imports to stargz-snapshotter. That patch adds `Initialized()` method that can be used for checking the readiness of Forgotten() metehod and we can avoid the issues like https://github.com/hanwen/go-fuse/issues/504 and https://github.com/containerd/stargz-snapshotter/issues/1594 .


cmd/go.mod
```diff
@@ -144,3 +144,5 @@ replace (
 	github.com/containerd/stargz-snapshotter/estargz => ../estargz
 	github.com/containerd/stargz-snapshotter/ipfs => ../ipfs
 )
+
+replace github.com/hanwen/go-fuse/v2 => github.com/ktock/go-fuse/v2 v2.1.1-0.20240611080711-f59a74eb644a
```
AkihiroSuda, line 148: Do you have a PR

ktock: Yes, there is a PR for this: https://github.com/hanwen/go-fuse/pull/520

PR #1773, opened 2024-08-21
Bump github.com/docker/cli to v27.1.2+incompatible

brandond:
Removes indirect dependency on `github.com/docker/docker`

@ktock I see that https://github.com/containerd/stargz-snapshotter/pull/1763 did the same thing and was closed, can I ask why? We'd really like to eliminate the indirect dependency on `github.com/docker/docker` since it is showing up in vuln scans when we bundle stargz-snapshotter into k3s and rke2.


brandond:
Edit: I see you're rolling these together in https://github.com/containerd/stargz-snapshotter/pull/1765 - I'll wait for that. Thanks!

PR #1874, opened 2024-11-25
fuse passthrough: fix PassthroughFd return wrong num to go-fuse

wswsmao:
In go-fuse, if the FilePassthroughFder interface is implemented, an attempt is made to register the fd to the kernel using addBackingID after an open operation. 
To avoid this operation in non-passthrough mode, we now return an exceptional value.


ktock (approved):
Thanks

PR #1893, opened 2024-12-10
Fix GC failure of CRI plugin

### LINKED ISSUE #1349: stargz-snapshotter uses up all available disk space
I have version 0.14.3 of the snapshotter installed on some EKS nodes, some of which have been running for around 16 days. They have started to run out of disk space and it seems the majority of this is consumed by `/var/lib/containerd-stargz-grpc/snapshotter/snapshots`.

Is there a way to prune/clean this up automatically?

ktock:
Fix #1349

CRI plugins fails to perform GC because it fails to find root direcotry of stargz snapshotter. 
Recent containerd added `proxy_plugins.stargz.exports.root` field to config.toml for solving this issue so this PR leverages that feature.

```
Dec 09 01:25:33 testkind-control-plane kubelet[2645]: E1209 01:25:33.155930    2645 kubelet.go:1467] "Image garbage collection failed once. Stats initialization may not have completed yet" err="get filesystem info: Failed to get the info of the filesystem with mountpoint: failed to get device for dir \"/var/lib/containerd/io.containerd.snapshotter.v1.stargz\": stat failed on /var/lib/containerd/io.containerd.snapshotter.v1.stargz with error: no such file or directory"
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,8 +71,6 @@ version = 2
   disable_snapshot_annotations = false
 ```
 
-**Note that `disable_snapshot_annotations = false` is required since containerd > v1.4.2**
-
 You can try our [prebuilt](/Dockerfile) [KinD](https://github.com/kubernetes-sigs/kind) node image that contains the above configuration.
 
 ```console
--- a/README.md
+++ b/README.md
@@ -62,6 +62,8 @@ version = 2
   [proxy_plugins.stargz]
     type = "snapshot"
     address = "/run/containerd-stargz-grpc/containerd-stargz-grpc.sock"
+  [proxy_plugins.stargz.exports]
+    root = "/var/lib/containerd-stargz-grpc/"
 
 # Use stargz snapshotter through CRI
 [plugins."io.containerd.grpc.v1.cri".containerd]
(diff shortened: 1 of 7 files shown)


docs/INSTALL.md
```diff
@@ -44,9 +44,13 @@ We assume that you are using containerd (> v1.4.2) as a CRI runtime.
     [proxy_plugins.stargz]
       type = "snapshot"
       address = "/run/containerd-stargz-grpc/containerd-stargz-grpc.sock"
+    [proxy_plugins.stargz.exports]
+      root = "/var/lib/containerd-stargz-grpc/"
 
   ```
 
+> NOTE: `root` field of `proxy_plugins` requires containerd >= v1.6.32 or v1.7.16 or v2.0.0
```
AkihiroSuda: Is this note necessary?

README.md
```diff
@@ -62,14 +62,17 @@ version = 2
...
+    root = "/var/lib/containerd-stargz-grpc/"
 
 # Use stargz snapshotter through CRI
 [plugins."io.containerd.grpc.v1.cri".containerd]
   snapshotter = "stargz"
   disable_snapshot_annotations = false
 ```
 
-**Note that `disable_snapshot_annotations = false` is required since containerd > v1.4.2**
+> NOTE1: `disable_snapshot_annotations = false` is required since containerd > v1.4.2
```
AkihiroSuda: You can just drop this note as containerd v1.4 and older releases have already reached EOL

docs/INSTALL.md
ktock: Thanks for the review. I think this is not necessary. I've just removed this note.

README.md
ktock: Removed this note.

AkihiroSuda (approved):

PR #1905, opened 2024-12-16
fusemanager: fix container fail after ttl timeout in detach mode

wswsmao:
In detach mode, when `containerd-stargz-grpc` exits normally, it sends an Unmount request to the fuse manager. Additionally, during the startup of `containerd-stargz-grpc`, the `restoreRemoteSnapshot` function cleans up previous mountpoints. If there are still running containers at this time, it can lead to issues when the TTL cache expires, resulting in abnormal behavior of the containers.

I considered several solutions:

1. The approach in the current PR, where users should restart `containerd-stargz-grpc` using `SIGKILL`, and then skip the cleanup step in `restoreRemoteSnapshot`.
2. Setting `ResolveResultEntryTTLSec` to an infinitely large value to leverage the TTL cache for ensuring the normal operation of containers. However, this would still lead to failures if the containers attempt to access uncached content.
3. Implementing a complex mechanism to determine if any running containers are using the mountpoints, and if so, skipping the cleanup.

After careful consideration, I have decided to proceed with the first approach.

### DIFF
--- a/docs/overview.md
+++ b/docs/overview.md
@@ -118,6 +118,18 @@ When upgrading the fuse manager, it's recommended to follow these steps:
 
 This ensures a clean upgrade without impacting running containers.
 
+### Important Considerations
+
+Before restarting the `containerd-stargz-grpc` process, it is essential to consider the state of any running containers.
+
+1. **When to Use SIGKILL** :
+
+If there are running containers, it is crucial to terminate the `containerd-stargz-grpc` process using `SIGKILL`. This approach prevents the normal shutdown sequence from attempting to clean up the mount points of the running containers, which could disrupt their availability. By using `SIGKILL`, you ensure that the process is forcefully terminated without affecting the ongoing operations of the containers.
+
+2. **When to Use SIGTERM** :
+
+If there are no running containers, you should use `SIGTERM` to terminate the `containerd-stargz-grpc` process. This allows the process to follow its normal shutdown sequence, ensuring that it properly cleans up resources and mount points.
+
 ## Registry-related configuration
 
 You can configure stargz snapshotter for accessing registries with custom configurations.
--- a/snapshot/snapshot.go
+++ b/snapshot/snapshot.go
@@ -73,6 +73,7 @@ type SnapshotterConfig struct {
 	asyncRemove                 bool
 	noRestore                   bool
 	allowInvalidMountsOnRestart bool
+	detach                      bool
 }
 
 // Opt is an option to configure the remote snapshotter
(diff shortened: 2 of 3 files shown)


snapshot/snapshot.go
```diff
@@ -736,14 +744,18 @@ func (o *snapshotter) checkAvailability(ctx context.Context, key string) bool {
...
-	mounts, err := mountinfo.GetMounts(nil)
-	if err != nil {
-		return err
-	}
-	for _, m := range mounts {
-		if strings.HasPrefix(m.Mountpoint, filepath.Join(o.root, "snapshots")) {
-			if err := syscall.Unmount(m.Mountpoint, syscall.MNT_FORCE); err != nil {
-				return fmt.Errorf("failed to unmount %s: %w", m.Mountpoint, err)
+	// In detach mode, rs is taken over by fusemanager,
+	// and there may be running containers, so we skip clean
```
ktock, line 748: No need to mention about fusemanager as snapshot/snapshot.go is agnostic about fusemanager.

snapshot/snapshot.go
```diff
@@ -736,14 +744,18 @@ func (o *snapshotter) checkAvailability(ctx context.Context, key string) bool {
...
+	if !o.detach {
```
ktock, line 749: Why not just retruning from this function?

docs/overview.md
```diff
@@ -118,6 +118,10 @@ When upgrading the fuse manager, it's recommended to follow these steps:
 
 This ensures a clean upgrade without impacting running containers.
 
+### Important Considerations
+
+Before restarting the `containerd-stargz-grpc` process, if there are running containers, it is crucial to use `SIGKILL` to terminate the `containerd-stargz-grpc` process. This approach prevents the normal shutdown sequence from attempting to clean up the mountpoints of the running containers, which could lead to disruptions in their availability. By using `SIGKILL`, you ensure that the process is forcefully terminated without affecting the ongoing operations of the containers.
```
ktock, line 123: Could you add a doc about when the user should use SIGTERM?

snapshot/snapshot.go
wswsmao: done

snapshot/snapshot.go
wswsmao: done，if o.detach will return directly

docs/overview.md
wswsmao: done

wswsmao:
all done

wswsmao:
@ktock

ktock (approved):
Thanks

PR #1920, opened 2025-01-06
Update gomod dependabot to use directories

djdongjin:
1. Move repeated `gomod` sections to a single one by using [`directories`](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/dependabot-options-reference#directories-or-directory--) instead of `directory`.
2. Add a few `groups`: update deps matching the `groups` pattern in a single PR.
  a. One benefit of this is dependabot will update the deps in all `go.mod`s, thus the CI can pass (otherwise it may fail due to requiring `go mod tidy`.
  b. I think a should be true regardless if `groups` is used or not. But from my experience, it seems dependabot only do this for `groups`ed deps, and still open different PRs for different modules if a dep is not part of a group.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -3,30 +3,32 @@ updates:
 
   # Automatic upgrade for go modules.
   - package-ecosystem: "gomod"
-    directory: "/"
+    directories:
+      - "/estargz"
+      - "/ipfs"
+      - "/"
+      - "/cmd"
     schedule:
       interval: "daily"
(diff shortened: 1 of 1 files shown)


ktock:
Could you rebase this on the latest main branch to enable CI?

djdongjin:
> Could you rebase this on the latest main branch to enable CI?

no problem, done.

ktock (approved):
Thanks

PR #1921, opened 2025-01-06
replace harshicorp/multierror with errors.Join

djdongjin:
As titled. The replacement is done by:

1. a `multierror` error (`allErr`) -> `var errs []error`
2. `multierror.Append(allErr, err)` -> `errs = append(errs, err)`
3. `return allErr` -> `return errors.Join(errs)`

And if `errs` is empty slice or if every `error` in `errs` are nil, `return errors.Join(errs)` will also return a nil error.

### DIFF
--- a/analyzer/fanotify/fanotify.go
+++ b/analyzer/fanotify/fanotify.go
@@ -17,6 +17,7 @@
 package fanotify
 
 import (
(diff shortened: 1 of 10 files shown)


ktock (approved):
Thanks

PR #1923, opened 2025-01-06
fuse passthrough: fix oom when running huge images

wswsmao:
During the testing of passthrough with large images, we observed that the containerd-stargz-grpc process was terminated. The reason for this issue is that the content to be merged is currently stored in bufPool, which poses a risk of memory exhaustion.

This commit modifies the implementation to directly write the chunks obtained from prefetchEntireFile to disk, thereby mitigating the risk of running out of memory.

### DIFF
--- a/fs/reader/reader.go
+++ b/fs/reader/reader.go
@@ -519,7 +519,7 @@ func (sf *file) GetPassthroughFd() (uintptr, error) {
 	// cache.PassThrough() is necessary to take over files
 	r, err := sf.gr.cache.Get(id, cache.PassThrough())
 	if err != nil {
-		if id, err = sf.prefetchEntireFile(); err != nil {
+		if err := sf.prefetchEntireFile(id); err != nil {
 			return 0, err
 		}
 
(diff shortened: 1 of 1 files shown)


ktock:
Could you rebase this on the latest main branch to enable CI?

wswsmao:
> Could you rebase this on the latest main branch to enable CI?

done, and this https://github.com/containerd/stargz-snapshotter/pull/1905 also has finished rebase and ready to be merged

fs/reader/reader.go
```diff
@@ -624,6 +625,31 @@ func (gr *reader) cacheData(ip []byte, cacheID string) {
...
+	file, ok := readerAt.(*os.File)
+	if !ok {
+		return fmt.Errorf("the cached ReaderAt is not of type *os.File")
+	}
+
+	if _, err := file.Seek(0, io.SeekEnd); err != nil {
+		return fmt.Errorf("failed to seek to end of file: %w", err)
+	}
+
+	if _, err := file.Write(ip); err != nil {
```
ktock, line 646: Blobs in the cache shouldn't be modified directly. The blob should be added the cache after the entire contents becoming available.

wswsmao: > Blobs in the cache shouldn't be modified directly. The blob should be added to the cache after the entire contents becoming available.

I have considered two potential solutions:

First, cache all chunks locally using cacheData, and then merge them into a single file by using cache.Get -> ReadAt -> file.Write. However, this approach incurs significant overhead and results in redundant files.

Limit the specifications of the images being used. Before calling combinedBuffer.Write, we can check for potential OOM risks. If such a risk is detected, we can immediately report an error and exit.

I was wondering if there are any other solutions you would recommend?

ktock: `cache.BlobCache.Add()` returns a writer and you can write chunks to that writer. Once the entire data is written to that writer, you can call `cache.Writer.Commit()` method then that blob is added to the cache. So can we use `cache.Writer` instead of `*os.File`?

wswsmao: done

fs/reader/reader.go
```diff
@@ -542,15 +542,24 @@ func (sf *file) GetPassthroughFd() (uintptr, error) {
...
-	combinedBuffer.Reset()
-	defer sf.gr.putBuffer(combinedBuffer)
+
+	w, err := sf.gr.cache.Add(entireCacheID)
+	if err != nil {
+		return fmt.Errorf("failed to create cache writer: %w", err)
+	}
+	defer w.Close()
+
+	seeker, ok := w.(io.Seeker)
```
ktock, line 558: Do we really need to seek? If it's guaranteed that we only append data to the cached blob, can we do this without seeking? Or are there other reasons to do seeking?

wswsmao: Oh, indeed, I made a small mistake, and I have fixed it, done

ktock (approved):
Thanks

PR #1941, opened 2025-01-15
Create a dependabot group for all go deps

djdongjin:
Dependabot seems to only update a dep in all go modoules, if it belongs to a group.

This new group will cover all deps that are not part of an existing group, so they get updated in all go.mod.

I think this will help avoid issues like https://github.com/containerd/stargz-snapshotter/pull/1939, and make dependabot work in most cases.


ktock (approved):
Thanks

Issue #1953, opened 2025-01-24
Implement Hardlink Feature for Cache Optimization and Data Deduplication


ChengyuZhu6:
# Description
I would like to propose the implementation of a hardlink feature in the caching mechanism to optimize memory usage, improve performance and , and save disk space.

# Background
The current caching system stores files in memory, which can lead to high memory usage, especially when dealing with large datasets. By utilizing hardlinks, we can reduce memory consumption and storage redundancy by allowing multiple references to the same file on disk without duplicating the file content.

# Design

## Key Components
1. HardlinkManager: Manages the creation, validation, and persistence of hardlinks.
- CreateLink: Attempts to create a hardlink for a given cache key.
- HasHardlink: Checks if a hardlink exists for a given key.
- Persist and Restore: Manages the persistence of hardlink metadata to disk and restores it on startup.
2. DirectoryCache: Implements the cache logic, including hardlink support.
- CreateHardlink: Invokes the HardlinkManager to create a hardlink.
- HasHardlink: Checks for the existence of a hardlink using the HardlinkManager.
3. Configuration: The EnableHardlink flag in the configuration determines whether hardlinking is enabled.

## Work Flow
```
[Start] 
   |
   v
[Initialize Cache]
   |
   v
[Check if Hardlinking is Enabled]
   |
   v
[Access Cached File] 
   |
   v
[Check if Hardlink Exists] -- No --> [Create Hardlink]
   |                                   |
  Yes                                  v
   |                             [Verify Hardlink]
   v                                   |
[Use Hardlink]                         v
   |                             [Rename to Final Location]
   v                                   |
[Persist Hardlink State] <-------------|
   |
   v
[Restore Hardlink State on Startup]
   |
   v
[End]
+-----------------------------+
```

1. Cache Write:
- When a file is added to the cache, the system checks if hardlinking is enabled.
- If enabled, it attempts to create a hardlink for the cached file.
2. Cache Read:
- When accessing a cached file, the system checks if a hardlink exists.
- If a hardlink exists, it uses the hardlink path to access the file.
Persistence:
- Hardlink metadata is periodically persisted to disk.
- On startup, the system restores hardlink metadata from disk.

# Benefits
- Reduced Memory Usage: By leveraging hardlinks, we can significantly decrease the memory footprint of the caching system.
- Improved Performance: Hardlinks allow for faster access to cached files, as they avoid the overhead of duplicating file data.
[...]

PR #1948, opened 2025-01-19
fs/remote: Refactor blob code to make it more modular

ChengyuZhu6:
This PR refactors the blob handling code to improve code organization. The changes include:

1. Extract chunk handling logic into separate functions:
   - prepareChunksForRead: Prepare chunks for reading
   - readFromCache: Handle cache reading
   - cacheChunkData: Handle chunk caching
   - handleSharedFetch: Handle shared fetch scenarios
   - copyFetchedChunks: Copy fetched chunks

2. Improve error handling:
   - Add more descriptive error messages
   - Consistent error wrapping using fmt.Errorf
   - Maintain same error handling behavior

3. Keep original behavior:
   - Maintain same data flow
   - Keep same concurrency control
   - No additional memory allocations
   - Same caching mechanism

4. Code organization improvements:
   - Single responsibility functions
   - Clear function names
   - Better code readability
   - Easier to maintain and test

PR #1954, opened 2025-01-24
Add hardlink support

ChengyuZhu6:
Propose the implementation of a hardlink feature in the caching mechanism to optimize memory usage, improve performance and save disk space.

Fixes: #1953 

Depends-on: #1948


cache/cache.go
```diff
@@ -281,81 +308,67 @@ func (dc *directoryCache) Add(key string, opts ...Option) (Writer, error) {
...
-	}
-	w := &writer{
-		WriteCloser: wip,
-		commitFunc: func() error {
-			if dc.isClosed() {
-				return fmt.Errorf("cache is already closed")
+	// Check hardlink first before creating new cache file
+	if dc.hlManager != nil {
+		if linkPath, exists := dc.hlManager.GetLink(key); exists {
+			if _, err := os.Stat(linkPath); err == nil {
```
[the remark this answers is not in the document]
ChengyuZhu6: I’m not sure I fully understand your point. Are you asking, 'Why do we need to check if a file with the same content has been hardlinked?' If that’s the case, let me explain: If a file has the same content, we wouldn’t need to create a new cache file; instead, we could simply hardlink to the existing one and use it. This approach would be beneficial for performance.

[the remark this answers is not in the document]
ChengyuZhu6: I think it would be useful if I can try to refactor the `CreateLink` function to use the `wipfile` function.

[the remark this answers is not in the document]
ChengyuZhu6: If I have misunderstood anything, please feel free to correct me.

cache/cache.go
```diff
@@ -205,8 +228,21 @@ func (dc *directoryCache) Get(key string, opts ...Option) (Reader, error) {
 		opt = o(opt)
 	}
 
+	// 1. Try to get from hardlink first
+	if dc.hlManager != nil {
+		if linkPath, exists := dc.hlManager.GetLink(key); exists {
+			if r, err := os.Open(linkPath); err == nil {
```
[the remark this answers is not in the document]
ChengyuZhu6: That's not related to the os.Open function. However, I think it's a good point, and I can reuse the `dc.cachePath(key)` function.

docs/hardlink.md
```diff
@@ -0,0 +1,59 @@
...
+- **Performance**: Evaluate the performance improvements and storage savings achieved by enabling hardlinking.
+
+## Troubleshooting
+
+- **Hardlink Not Created**: Ensure that the `EnableHardlink` option is set to `true` and that the configuration file is correctly loaded by `stargz-snapshotter`.
+- **File Access Issues**: Check the logs for any errors related to hardlink creation or access. Ensure that the underlying filesystem supports hardlinks.
+
+## Conclusion
+
+Enabling hardlinking in `stargz-snapshotter` can significantly optimize storage and improve performance. By following the steps outlined in this guide, you can leverage this feature to enhance your containerized environments.
```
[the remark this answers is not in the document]
ChengyuZhu6: Are you referring to specific numbers regarding performance or storage savings?

cache/cache.go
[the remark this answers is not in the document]
ChengyuZhu6: Got your point. I'll refactor my code.

ChengyuZhu6:
> Needs rebase

Done.

ChengyuZhu6:
Just retrigger ci, no code change.

ChengyuZhu6:
@ktock I conducted experiments with several basic images, converting them to the estargz format and running them in containers with a simple 'echo "hello"' command. These tests used only background threads of stargz to pull images to the local machine. By measuring the overall memory and disk usage, I observed that implementing hardlinks resulted in a 20-30% reduction in both memory consumption and disk space requirements.

image
image

ChengyuZhu6:
When working with different versions of an image, implementing hardlinks can achieve a memory and disk deduplication effect of nearly 50%. I verified this by conducting tests with various development versions of Golang images.

image

ChengyuZhu6:
kindly ping @ktock

cache/cache.go
```diff
@@ -166,15 +178,24 @@ func NewDirectoryCache(directory string, config DirectoryCacheConfig) (BlobCache
...
+		wipDirectory:   wipdir,
+		bufPool:        bufPool,
+		direct:         config.Direct,
+		enableHardlink: config.EnableHardlink,
+		syncAdd:        config.SyncAdd,
+	}
+
+	// Initialize hardlink manager if enabled
+	if config.EnableHardlink {
+		hlManager, enabled := InitializeHardlinkManager(filepath.Dir(filepath.Dir(directory)), config.EnableHardlink)
```
ktock: can you deduplicate checks of `config.EnableHardlink`

cache/cache.go
```diff
@@ -123,6 +127,14 @@ func PassThrough() Option {
 	}
 }
 
+// ChunkDigest option allows specifying a chunk digest for the cache
+func ChunkDigest(digest string) Option {
```
ktock, line 140: why not use digest.Digest

cache/cache.go
```diff
@@ -229,10 +259,21 @@ func (dc *directoryCache) Get(key string, opts ...Option) (Reader, error) {
 		}
 	}
 
+	// First try regular file path
+	filepath := BuildCachePath(dc.directory, key)
+
+	// Check hardlink manager for existing digest file
+	if dc.hlManager != nil && opt.chunkDigest != "" {
```
ktock: no need to check `dc.hlManager.IsEnabled()`?
anyway maintaining `dc.hlManager != nil && dc.hlManager.IsEnabled() && opt.chunkDigest != "" ` check everywhere is hard, so this should be defined as a function somewhere with a clear explanation.

fs/remote/blob.go
```diff
@@ -89,6 +90,7 @@ func makeBlob(fetcher fetcher, size int64, chunkSize int64, prefetchChunkSize in
 		checkInterval:     checkInterval,
 		resolver:          r,
 		fetchTimeout:      fetchTimeout,
+		blobDigest:        blobDigest,
```
ktock: why?

cmd/containerd-stargz-grpc/db/reader.go
```diff
@@ -1156,3 +1160,39 @@ func (cr *countReader) Read(p []byte) (n int, err error) {
...
+			if err != nil {
+				return err
+			}
+			if len(chunks) > 0 {
+				if isChunk {
+					digest = chunks[0].chunkDigest
+				} else {
+					digest = chunks[0].fileDigest
+				}
+			}
```
ktock: How the caller can get following chunks digests?

fs/layer/layer.go
```diff
@@ -518,14 +520,22 @@ func (l *layer) prefetch(ctx context.Context, prefetchSize int64) error {
...
+		digest, err = l.verifiableReader.Metadata().GetDigest(rootID, false)
+		if err != nil {
+			return fmt.Errorf("failed to get digest of prefetch landmark: %w", err)
+		}
 	}
 
 	// Fetch the target range
 	downloadStart := time.Now()
-	err := l.blob.Cache(0, prefetchSize)
+	err := l.blob.Cache(0, prefetchSize, remote.WithCacheOpts(cache.ChunkDigest(digest)))
```
ktock: Doesn't it result in `blob.Cache()` querying all chunks (including chunks[1:]) with `chunks[0]`, according to the current db change in this patch?

ChengyuZhu6: removed

cmd/containerd-stargz-grpc/db/reader.go
ChengyuZhu6: removed

fs/remote/blob.go
ChengyuZhu6: removed.

cache/cache.go
ChengyuZhu6: Done.

cache/cache.go
ChengyuZhu6: Done.

cache/cache.go
ChengyuZhu6: Done.

ChengyuZhu6:
@ktock

metadata/metadata.go
```diff
@@ -78,7 +78,7 @@ type Reader interface {
 }
 
 type File interface {
-	ChunkEntryForOffset(offset int64) (off int64, size int64, dgst string, ok bool)
+	ChunkEntryForOffset(offset int64) (off int64, size int64, dgst string, fileDigest string, ok bool)
```
ktock: This should be smtg like `Digest() digest.Digest` method.

fs/layer/layer.go
```diff
@@ -521,6 +522,9 @@ func (l *layer) prefetch(ctx context.Context, prefetchSize int64) error {
 	} else if prefetchSize > l.blob.Size() {
 		// adjust prefetch size not to exceed the whole layer size
 		prefetchSize = l.blob.Size()
+		if err != nil {
+			return fmt.Errorf("failed to get digest of prefetch landmark: %w", err)
+		}
```
ktock: why?

cache/utils.go
```diff
@@ -0,0 +1,40 @@
...
+
+//revive:enable:exported
+
+// WipFile creates a temporary file in the given directory with the given key pattern
+func WipFile(wipDirectory string, key string) (*os.File, error) {
+	if err := os.MkdirAll(wipDirectory, 0700); err != nil {
+		return nil, fmt.Errorf("failed to create wip directory: %w", err)
+	}
+	return os.CreateTemp(wipDirectory, key+"-*")
+}
```
ktock: They don't need to be exported.

cache/cache.go
```diff
@@ -205,9 +229,15 @@ func (dc *directoryCache) Get(key string, opts ...Option) (Reader, error) {
...
 
+	// Try to get from memory cache
 	if !dc.direct && !opt.direct {
-		// Get data from memory
-		if b, done, ok := dc.cache.Get(key); ok {
+		// Try memory cache for digest or key
+		cacheKey := key
+		if shouldUseDigestCacheKey(dc.hlManager, opt.chunkDigest) {
+			cacheKey = opt.chunkDigest
+		}
```
ktock: `key` can be an arbitrary string so this can cause a key conflict, can't it? To prevent this, the key for `dc.cache` should be prefixed with the type of usage (e.g. `raw:` for the raw key and `digest:` for the digest)

cache/cache.go
```diff
@@ -173,8 +185,18 @@ func NewDirectoryCache(directory string, config DirectoryCacheConfig) (BlobCache
...
 		bufPool:      bufPool,
 		direct:       config.Direct,
+		syncAdd:      config.SyncAdd,
 	}
-	dc.syncAdd = config.SyncAdd
+
+	// Initialize hardlink manager if enabled
+	if config.EnableHardlink {
+		var err error
+		dc.hlManager, err = InitializeHardlinkManager(filepath.Dir(filepath.Dir(directory)))
```
ktock: This package should not access to the parent directory of the passed root ditrectory path. If this is a global service using the directory outside of the passed root directory, this should be instantiated outside of this package.

cache/cache.go
```diff
@@ -229,10 +259,21 @@ func (dc *directoryCache) Get(key string, opts ...Option) (Reader, error) {
...
+	// First try regular file path
+	filepath := BuildCachePath(dc.directory, key)
+
+	// Check hardlink manager for existing digest file
+	if shouldUseDigestCacheKey(dc.hlManager, opt.chunkDigest) {
+		if digestPath, exists := dc.hlManager.ProcessCacheGet(key, opt.chunkDigest, opt.direct); exists {
+			log.L.Debugf("Using existing file for digest %q instead of key %q", opt.chunkDigest, key)
+			filepath = digestPath
+		}
+	}
```
ktock, line 270: Doesn't this need to create a hardlink? What happens if the source file is deleted before opening it?

cache/hardlink.go
```diff
@@ -0,0 +1,519 @@
...
+			}
+			hm.mu.Unlock()
+		case <-hm.persistDone:
+			return
+		}
+	}
+}
+
+// Close stops background goroutines and persists data
+func (hm *HardlinkManager) Close() error {
```
ktock: This should be called when shutting down the snapshotter.

cache/hardlink.go
```diff
@@ -0,0 +1,519 @@
...
+}
+
+// HardlinkManager manages digest-to-file mappings and key-to-digest mappings
+type HardlinkManager struct {
+	root            string
+	hlDir           string
+	mu              sync.RWMutex
+	digestToKeys    map[string]*ChunkDigestMapping // Maps chunkdigest to its associated keys
+	keyToDigest     map[string]string              // Reverse index: maps key to its chunkdigest
+	digestToFile    map[string]string              // Maps chunkdigest directly to file path
```
ktock: Are they scalable? Can they be huge maps if the node contains lots of images?

cache/hardlink.go
```diff
@@ -0,0 +1,519 @@
...
+	hm.persistTicker.Stop()
+	hm.cleanupTicker.Stop()
+	close(hm.persistDone)
+	close(hm.cleanupDone)
+	// Final persist of any remaining changes
+	return hm.persist()
+}
+
+// HardlinkCapability represents a cache that supports hardlinking
+type HardlinkCapability interface {
```
ktock: Where is this used?

ktock:
CI failure in `Optimize` will be fixed in #2094.

cache/cache.go
ChengyuZhu6: If it’s deleted before open/Get: GetLink stats the path, sees it missing, removes the digest from the LRU (OnEvicted also drops DigestToKeys/KeyToDigest), and returns "", false. Callers (ProcessCacheGet/Add) treat it as a miss and proceed with normal file creation (no hardlink).
If it’s deleted between GetLink and os.Link: CreateLink’s os.Link fails; the error bubbles up, caller logs debug and falls back to normal creation.

cache/utils.go
ChengyuZhu6: Removed.

fs/layer/layer.go
ChengyuZhu6: Removed.

cache/hardlink.go
ChengyuZhu6: I have changed the code to use LRUCache for digestToFile

cache/hardlink.go
ChengyuZhu6: Removed

cache/hardlink.go
ChengyuZhu6: I have removed the persistence operations to make the code clear. The code would not affect the cache logic. On miss or failure, code falls back to creating a new file.

ChengyuZhu6: So we don't need Close() function.

cache/cache.go
ChengyuZhu6: There is no need to persist hardlink metadata at present, so hardlinkmanager does not need to create a directory for storing hardlink data.

ChengyuZhu6: I've moved the hardlink code to hardlink package and initialize it in main.go

cache/cache.go
ChengyuZhu6: Good idea. I've changed the code.

metadata/metadata.go
ChengyuZhu6: @ktock I think it may not necessary. The file-level digest(fileDigest) is only used in the passthrough path; the normal path relies on per‑chunk digests (strings). These would both call `cache.ChunkDigest` to pass the digest.

hardlink/hardlink.go
```diff
@@ -0,0 +1,206 @@
...
+
+// RegisterDigestFile registers a file as the primary source for a digest
+func (hm *Manager) RegisterDigestFile(chunkdigest string, filePath string) error {
+	if _, err := os.Stat(filePath); err != nil {
+		return fmt.Errorf("file does not exist at %q: %w", filePath, err)
+	}
+	_, done, _ := hm.digestToFile.Add(chunkdigest, filePath)
+	done()
+	return nil
+}
```
ktock: Now GetLink returns a source file registered to `digestToFile` through `(*directoryCache).Add()` -> `(*Manager).RegisterDigestFile()`, doesn't it? If so, what's the point of creating a hardlink via `CreateLink`?

hardlink/hardlink.go
```diff
@@ -0,0 +1,206 @@
...
+
+// GetDigestForKey returns the digest mapped to the given key, if any.
+func (hm *Manager) GetDigestForKey(key string) (string, bool) {
+	hm.mu.RLock()
+	defer hm.mu.RUnlock()
+	d, ok := hm.keyToDigest[key]
+	return d, ok
+}
```
ktock: This isn't used other than tests so should be removed from this file.

fs/reader/reader.go
```diff
@@ -509,18 +510,22 @@ func (sf *file) GetPassthroughFd(mergeBufferSize int64, mergeWorkerCount int) (u
...
-		chunkOffset, chunkSize, digestStr, ok := sf.fr.ChunkEntryForOffset(offset)
+		chunkOffset, chunkSize, digestStr, digest, ok := sf.fr.ChunkEntryForOffset(offset)
 		if !ok {
 			break
 		}
 		// Check if any chunk size exceeds merge buffer size to avoid bounds out of range
 		if chunkSize > mergeBufferSize {
 			hasLargeChunk = true
 		}
+		if firstChunkOffset == 0 {
```
ktock: What's the point of this check? firstChunkOffset looks like always 0. Anyway this variable should be removed as it's confusing, but not in this PR. cc @wswsmao

cache/cache.go
```diff
@@ -273,7 +314,12 @@ func (dc *directoryCache) Get(key string, opts ...Option) (Reader, error) {
 	return &reader{
 		ReaderAt: file,
 		closeFunc: func() error {
-			_, done, added := dc.fileCache.Add(key, file)
+			cacheKey := cacheKeyPrefixRaw + key
+			if shouldUseDigestCacheKey(dc.hlManager, opt.chunkDigest) {
+				cacheKey = cacheKeyPrefixDigest + opt.chunkDigest
+			}
```
ktock: This appears 2 times in this function so can you deduplicate this?

cache/cache.go
```diff
@@ -293,88 +339,78 @@ func (dc *directoryCache) Add(key string, opts ...Option) (Writer, error) {
...
+
+			if shouldUseDigestCacheKey(dc.hlManager, opt.chunkDigest) {
+				if err := dc.hlManager.RegisterDigestFile(opt.chunkDigest, targetPath); err != nil {
+					return fmt.Errorf("failed to register digest file: %w", err)
+				}
+
+				internalKey := dc.hlManager.GenerateInternalKey(dc.directory, key)
+				if err := dc.hlManager.MapKeyToDigest(internalKey, opt.chunkDigest); err != nil {
+					return fmt.Errorf("failed to map key to digest: %w", err)
+				}
```
ktock: This looks like an internal logic of the manager so this should also be done in RegisterDigestFile.

cache/cache.go
```diff
@@ -178,8 +195,14 @@ func NewDirectoryCache(directory string, config DirectoryCacheConfig) (BlobCache
 		bufPool:      bufPool,
 		direct:       config.Direct,
 		fadvDontNeed: config.FadvDontNeed,
+		syncAdd:      config.SyncAdd,
+	}
+
+	// Initialize hardlink manager if enabled
+	if config.EnableHardlink {
+		dc.hlManager = hardlink.GetGlobalManager()
```
ktock: This module is also used by tools other than stargz snapshotter (e.g. stargz store) so please do not assume something global. The manager instance should be passed via a flag.

cmd/containerd-stargz-grpc/db/reader.go
```diff
@@ -450,14 +451,17 @@ func (r *reader) initNodes(tr io.Reader) error {
 					wantNextOffsetID = append(wantNextOffsetID, id)
 				}
 
+				if ent.Type == "reg" {
+					fileDigest = ent.Digest
+				}
```
ktock: `ChunkEntryForOffset` API literally intends to return a TOCEntry information of the chunk so this shouldn't be done.

metadata/memory/reader.go
```diff
@@ -245,18 +245,18 @@ type file struct {
...
+		return 0, 0, "", "", false
 	}
 	dgst = e.Digest
 	if e.ChunkDigest != "" {
 		// NOTE* "reg" also can contain ChunkDigest (e.g. when "reg" is the first entry of
 		// chunked file)
 		dgst = e.ChunkDigest
 	}
-	return e.ChunkOffset, e.ChunkSize, dgst, true
+	return e.ChunkOffset, e.ChunkSize, dgst, e.Digest, true
```
ktock: This looks inconsistent against the change you've added to ~~`cmd/containerd-stargz-grpc/db/db.go`~~ `cmd/containerd-stargz-grpc/db/reader.go` which propagated the first chunk's digest field to other chunks as well.

hardlink/hardlink.go
ktock: Or shouldn't `digestToFile` be a set of hardlinks on a directory + periodical GC based on the nlink?

fs/reader/reader.go
wswsmao: remove firstChunkOffset in https://github.com/containerd/stargz-snapshotter/pull/2112

hardlink/hardlink.go
ChengyuZhu6: Done.

cmd/containerd-stargz-grpc/db/reader.go
ChengyuZhu6: removed.

metadata/memory/reader.go
ChengyuZhu6: Removed.

cache/cache.go
ChengyuZhu6: Done.

cache/cache.go
ChengyuZhu6: Done.

cache/cache.go
ChengyuZhu6: Done.

hardlink/hardlink.go
ChengyuZhu6: Done.

metadata/metadata.go
ChengyuZhu6: Removed.

hardlink/hardlink.go
```diff
@@ -0,0 +1,218 @@
...
+
+func (hm *Manager) cleanup(canonicalPath string) {
+	stat, err := os.Stat(canonicalPath)
+	if err != nil {
+		return
+	}
+
+	if sys, ok := stat.Sys().(*unix.Stat_t); ok {
+		if sys.Nlink == 1 {
```
nibble0x2, line 81: Hi @ChengyuZhu6 ,

Really like the idea and see lots of benefits from this... have few Qs around the implementation - 

1) when `Nlink` becomes 1? as far as I understand - stargz never deletes its temp dir cache so Nlink will never become 1? https://github.com/containerd/stargz-snapshotter/blob/main/fs/layer/layer.go#L231 

2) while evicting at `hardlink_manager` side - what if cleanup never goes through i.e Nlink > 1 then these files will never get cleaned up as we are clearing LRU entries?

3) what's the purpose of DS  `digestToKeys` and `keyToDigest` ? as I see `#Get` completely relies on `chunkDigest` to retrieve the link

4) does hardlink_manager recover the existing state during start up? let's say stargz was running fine and it had to restart - should it populate `digestToFile` during initialization for existing hardlinks?

Thanks!

PR #1967, opened 2025-02-07
Add docs about how to use stargz-snapshotter on Lima

ktock:


docs/lima.md
```diff
@@ -0,0 +1,47 @@
+# Using stargz-snapshotter on Lima
+
+This document describes how to get started with stargz-snapshotter on [Lima](https://github.com/lima-vm/lima).
```
AkihiroSuda, line 3: Needs a brief explanation of Lima, otherwise readers won't get the point of this document.

docs/lima.md
```diff
@@ -0,0 +1,47 @@
...
+>>>
+```
+
+## Use Stargz Snapshotter as the default snapshotter
+
+nerdctl recognizes an environment variable `CONTAINERD_SNAPSHOTTER` for the snapshotter to use.
+You can add this environment variable to the VM by configuring Lima config as shown in the following:
+
+```
+$ cat <<EOF >> ~/.lima/_config/override.yaml
```
AkihiroSuda, line 30: It should be noted that `override.yaml` applies to all the instances of Lima

ktock:
Thanks for the review. Fixed the docs based on the comments.

AkihiroSuda (approved):
Thanks

PR #1987, opened 2025-02-20
concurrent file merging in passthrough mode to reduce initial pull time

wswsmao:
In passthrough mode, the initial pull of an image requires merging chunks into a file. This process can be time-consuming, especially for large files.

This pr optimizes the merging process by introducing two configuration options: `merge_buffer_size` and `merge_worker_count`. By concurrently reading chunks and caching them for batch writing, we enhance the performance of the initial image pull in passthrough mode.

The principle of concurrent reading is illustrated in the diagram below. 
The `merge_worker_count` corresponds to the number of workers in the diagram, while `merge_buffer_size` corresponds to the size of the buffer in the diagram. Before concurrent reading begins, the buffer area is divided, and each worker is assigned a specific region. This approach allows all workers to concurrently read the contents of the chunks into the buffer.
image

PR #2012, opened 2025-03-12
Add buffer validation and tests for passthrough concurrent file merging

wswsmao:
This PR addresses the outstanding issue discussed in [this comment](https://github.com/containerd/stargz-snapshotter/pull/1987#discussion_r1988314111). The following changes are introduced:

1. **New `checkHoles` method**
   During the `processBatchChunks`, we now record the starting position and the actual size of the buffer read each time. This information is then validated by the `checkHoles` method to ensure there are no gaps or overwrites in the buffer.
2. **Validation after `w.Write(buffer)`**
   We also added validation for the return value after `w.Write(buffer)` to prevent any unknown write errors.
3. **Supplementary test cases**
   Additional test cases have been included to cover three scenarios: normal batch processing with validation, batch processing with unhandled chunks, and errors in index calculations.

### DIFF
--- a/fs/reader/reader.go
+++ b/fs/reader/reader.go
@@ -32,6 +32,7 @@ import (
 	"io"
 	"os"
 	"runtime"
+	"sort"
 	"sync"
 	"time"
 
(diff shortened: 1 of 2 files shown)


ktock (approved):
Thanks

PR #2041, opened 2025-04-25
CI: Enable actions/attest-build-provenance@v2

takumin:
https://github.com/actions/attest-build-provenance

https://github.blog/changelog/2024-06-25-artifact-attestations-is-generally-available/


takumin:
Sorry, I was planning to do a test release, but I've become busy so I will have to cancel it.
Thank you for always providing us with great software.

PR #2068, opened 2025-06-25
fix panic when chunk size exceeds merge buffer size

wswsmao:
Add a check to ensure that chunkSize does not exceed mergeBufferSize. 
If chunkSize is greater than mergeBufferSize, return an error to prompt the user to update the config or adjust the image chunk size. 
This prevents potential slice bounds out of range panics during merging.


fs/reader/reader.go
```diff
@@ -516,6 +516,12 @@ func (sf *file) GetPassthroughFd(mergeBufferSize int64, mergeWorkerCount int) (u
 		if !ok {
 			break
 		}
+		// Check if chunk size exceeds merge buffer size to prevent slice bounds out of range panic
+		// when mergeBufferSize is smaller than chunkSize, report error to suggest user to
+		// update config or adjust image chunk size
+		if chunkSize > mergeBufferSize {
```
ktock, line 521: why not dynamically allocate?

wswsmao: Dynamic allocation also carries risks. For example, in environments with limited memory, if the chunk size is large and multiple chunks are being downloaded concurrently, it could lead to OOM (out-of-memory) issues—similar to the problem addressed in [this PR](https://github.com/containerd/stargz-snapshotter/pull/1923).

The purpose of making mergeBufferSize configurable is to allow users to set an acceptable buffer size based on their own environment’s resources and the characteristics of their images.

ktock: >   allow users to set an acceptable buffer size based on their own environment’s resources and the characteristics of their images.

It's better to automatically control (reduce or disable) the concurrency not to exceed the buffer limit. Snapshotter should not fail depending on the image size.

wswsmao: done， when chunkSize > mergeBufferSize,  it will automatically fall back to sequential processing using the legacy approach.

PR #2077, opened 2025-07-08
Make graceful restarting available in wider configurations

ktock:
Depends on  #2076
Following up https://github.com/containerd/stargz-snapshotter/pull/1892

When stargz snapshotter receives SIGINT, it perform the graceful exit with explicitly unmounting the mountpoints and releasing the associated resources.

This PR makes this avilablable also in the following situations

- when FUSE manager is killed by SIGINT. The user can use SIGTERM and SIGINT for different porpose
  - SIGTERM: Usable for restarting only containerd-stargz-grpc (e.g. for reloading configuration). No need to unmount snapshots because they are managed by the FUSE manager.
  - SIGINT: Usable when the user wants to use the graceful exit feature of containerd-stargz-grpc because the user will kill the FUSE manager for node restart, etc.
- when containerd-stargz-grpc is killed by SIGTERM (systemd's default [KillSignal](https://www.freedesktop.org/software/systemd/man/latest/systemd.kill.html#KillSignal=))

This PR also adds docs about how to gracefully kill and restart Stargz Snapshotter under different configurations.

### DIFF
--- a/cmd/containerd-stargz-grpc/main.go
+++ b/cmd/containerd-stargz-grpc/main.go
@@ -165,7 +165,7 @@ func main() {
 		if !filepath.IsAbs(fmAddr) {
 			log.G(ctx).WithError(err).Fatalf("fuse manager address must be an absolute path: %s", fmAddr)
 		}
-		err := fusemanager.StartFuseManager(ctx, fmPath, fmAddr, filepath.Join(*rootDir, "fusestore.db"), *logLevel, filepath.Join(*rootDir, "stargz-fuse-manager.log"))
+		managerNewlyStarted, err := fusemanager.StartFuseManager(ctx, fmPath, fmAddr, filepath.Join(*rootDir, "fusestore.db"), *logLevel, filepath.Join(*rootDir, "stargz-fuse-manager.log"))
 		if err != nil {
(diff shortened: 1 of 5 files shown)


fs/layer/layer.go
```diff
@@ -110,6 +110,10 @@ type Layer interface {
 	// Done releases the reference to this layer. The resources related to this layer will be
 	// discarded sooner or later. Queries after calling this function won't be serviced.
 	Done()
+
+	// Close is the same as Done. But this evicts the resources related to this Layer immediately.
+	// This can be used for cleaning up resources on unmount.
+	Close()
```
AkihiroSuda: Why not use https://pkg.go.dev/io#Closer

script/integration/test.sh
```diff
@@ -81,7 +81,7 @@ cat <<EOF > "${TMP_CONTEXT}/Dockerfile"
 FROM ${INTEGRATION_BASE_IMAGE_NAME}
 
 RUN apt-get update -y && \
-    apt-get --no-install-recommends install -y iptables jq && \
+    apt-get install -y iptables jq netcat && \
```
AkihiroSuda: Why drop `--no-install-recommends`

AkihiroSuda (approved):

PR #2084, opened 2025-07-15
Add docs and tests for Transfer Service

ktock:
Transfer Service is enabled in CRI by default since containerd v2.1. However, when the CRI plugin finds `disable_snapshot_annotations = false` in containerd's config.toml, it doesn't enable Transfer Service and falls back to the traditional client-library-based pulling. (c.f. [docs](https://github.com/containerd/containerd/blob/v2.1.0/docs/cri/config.md#configuration-differences-and-automatic-fallback-to-local-mode)). So if user preserve `disable_snapshot_annotations = false` in the config, the user can expect the traditional (non-Transfer-Service) behaviour without breaking lazy pulling behaviour.

If the user wants to enable Transfer Service, additional configuration is needed and this commit adds a doc about it.

Tests are also added for continuously checking if stargz snapshotter works with enabling Transfer Service.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -56,6 +56,7 @@ jobs:
         builtin: ["true", "false"]
         metadata-store: ["memory", "db"]
         fuse-passthrough: ["true", "false"]
+        transfer-service: ["true", "false"]
         exclude:
         - buildargs: ""
           builtin: "true"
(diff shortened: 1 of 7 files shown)


docs/transfer.md
```diff
@@ -0,0 +1,99 @@
+# Enabling Stargz Snapshotter With Transfer Service
+
+Transfer Service is a containerd component which is used for image management in contianerd (e.g. pulling and pushing images).
+For details about Transfer Service, refer to [the official document in the containerd repo](https://github.com/containerd/containerd/blob/6af7c07905a317d4c343a49255e2392f4c8569f9/docs/transfer.md).
+
+To use Stargz Snapshotter on containerd with enabling Transfer Service, additional configurations is needed.
+
+## When is Transfer Service enabled?
```
AkihiroSuda: "When" sounds awkward here
```suggestion
## Availability of Transfer Service
```


Anyway you can assume that the service is always available as containerd v1.6 will reach EOL soon

Issue #2093, opened 2025-07-25
Double pagecache consumption and solution


wswsmao:
Hi, 
currently,  there is a double pagecache issue:

1. **FUSE daemon level**: When stargz-snapshotter reads fscache files, data gets loaded into kernel pagecache
2. **Container process level**: When container processes read files through FUSE, the same data gets loaded into kernel pagecache again

The `Direct` mode only disables user-space memory cache but still uses regular `os.Open()` without `O_DIRECT` flags. This means backend fscache files are still cached in kernel pagecache, leading to double memory consumption.

There are two solutions available, and we can use a configuration option to decide whether to enable.

Option 1: Add posix_fadvise in closeFunc
```
// In (dc *directoryCache) Get method, modify closeFunc:
closeFunc: func() error {
    if err := posix_fadvise(file.Fd(), 0, 0, POSIX_FADV_DONTNEED); err != nil {
        log.Warnf("failed to drop pagecache: %v", err)
    }

    if opt.passThrough {
        return nil
    }
    // Add posix_fadvise to drop pagecache
    return file.Close()
}
```

Option 2: Use O_DIRECT for file operations
```
// In (dc *directoryCache) Get method, use O_DIRECT:
file, err := os.OpenFile(dc.cachePath(key), os.O_RDONLY|syscall.O_DIRECT, 0)
```
However, this approach may incur a performance overhead.

Are these two solutions acceptable? I can submit a PR to implement the one that makes the most sense.

Best regards,
Abush Wang

PR #2095, opened 2025-07-28
fix double pagecache consumption

wswsmao:
Fixes https://github.com/containerd/stargz-snapshotter/issues/2093


cache/cache.go
```diff
@@ -16,6 +16,13 @@
 
 package cache
 
+/*
+#include <fcntl.h>
+#include <unistd.h>
+#include <errno.h>
+*/
+import "C"
```
ktock: stargz snapshotter is imported from other projects as well so cgo should be avoided.

fs/config/config.go
```diff
@@ -139,6 +139,9 @@ type DirectoryCacheConfig struct {
 
 	// Direct disables on-memory data cache. Default is true for saving memory usage.
 	Direct bool `toml:"direct" default:"true" json:"direct"`
+
+	// FadvDontNeed forcefully clean fscache pagecache for saving memory. Default is false.
+	FadvDontNeed bool `toml:"fadv_dontneed" json:"fadv_dontneed"`
```
ktock, line 144: performance numbers?

wswsmao: > performance numbers?

Could you please clarify if you would like me to provide additional performance metrics or if you're suggesting that I should include a note on potential performance impacts in the documentation? Forgive me, I'm not entirely sure I understood this comment.

ktock: Sorry for the unclear comment, I meant, how big can we expect the memory saving improvement from this change in the actual workloads?

cache/cache.go
wswsmao: done

fs/config/config.go
wswsmao: > Sorry for the unclear comment, I meant, how big can we expect the memory saving improvement from this change in the actual workloads?

In theory, this approach can save up to 50% of memory.
I created a test image containing a 1GB file. The container under test runs a `sha512sum` operation on this file:

```
nerdctl run --rm -it --snapshotter stargz $REGISTRY_NAME/$IMAGE_NAME:alpine-1G-esgz sha512sum /bigfile
```

Meanwhile, in another terminal, I monitored the system's memory information:

```
watch -n 1 'cat /proc/meminfo | grep -E "Cached"'
```

Here are the test results, showing a reduction close to 50%:

| fadv\_dontneed | before  | after     | Cached    |
| ---------------- | --------- | ----------- | ----------- |
| disable        | 852,368 | 3,074,640 | 2,222,272 |
| enable         | 852,060 | 2,022,852 | 1,170,792 |

cache/cache.go
wswsmao: use unix instead

ktock (approved):
Thanks

Issue #1760, opened 2024-08-09
Remove `testing` dependency from public dependencies


howardjohn:
The `estargz/testutil.go` and similar files leak the stdlib `testing` dependency into any production code that depends on these utilities. It would be ideal if the testing stuff was kept isolated from the core code to minimize transitive dependencies. This isn't a huge deal but still nice to have IMO.

I took a briefly look and the way things are structured makes it not trivial to do so

PR #2098, opened 2025-07-30
Remove `testing` dependency from code outside tests

rosstimothy:
The various `testutil` packages reduce duplication, but exist outside of a `_test.go` package and they, along with all of their dependencies are not pruned via dead code elimination which causes #1760. To cut the dependency without refactoring the tests, this creates a local interface which covers the subset of the `testing.T` methods used by the test helpers. 

Closes https://github.com/containerd/stargz-snapshotter/issues/1760.


ktock:
Thanks, please add the signed-off-by line to you commit message. The failure in `Tests/Optimize` isn't related to this patch and will be fixed in #2094 .

PR #2113, opened 2025-08-25
ctr-remote: add prefetch-list option

wswsmao:
The optimize command requires a running container; however, this isn’t always feasible—for example, when the container needs GPU hardware while the service that performs image format conversion (e.g. a webhook) runs in an environment without GPUs. In such situations, users can collect the required dependency files themselves and provide them as input; the --prefetch-list option allows the format conversion to be carried out in the constrained environment.

### DIFF
--- a/cmd/ctr-remote/commands/optimize.go
+++ b/cmd/ctr-remote/commands/optimize.go
@@ -17,6 +17,7 @@
 package commands
 
 import (
+	"bufio"
 	"compress/gzip"
 	"context"
 	"encoding/json"
@@ -25,6 +26,7 @@ import (
 	"io"
 	"os"
 	"os/signal"
+	"strings"
 	"time"
 
 	containerd "github.com/containerd/containerd/v2/client"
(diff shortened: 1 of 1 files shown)


cmd/ctr-remote/commands/optimize.go
```diff
@@ -248,11 +259,121 @@ func writeContentFile(ctx context.Context, client *containerd.Client, dgst diges
...
+	}
+
+	if err := scanner.Err(); err != nil {
+		return nil, fmt.Errorf("failed to read prefetch list file: %w", err)
+	}
+
+	return paths, nil
+}
+
+func analyzePrefetchList(ctx context.Context, client *containerd.Client, srcRef string, prefetchPaths []string) (digest.Digest, map[digest.Digest][]estargz.Option, func(converter.ConvertFunc) converter.ConvertFunc, error) {
```
ktock: A large portion of this function looks the same as a part of `analyze` function (https://github.com/containerd/stargz-snapshotter/blob/ec299cc621c3dce3dd4b5f97dcd9c0583625e5af/cmd/ctr-remote/commands/optimize.go#L304-L360) except the manual creation of the record. So can you  split this part to a common helper function so that it can be used for the both flags?

wswsmao: done

cmd/ctr-remote/commands/optimize.go
```diff
@@ -248,89 +259,67 @@ func writeContentFile(ctx context.Context, client *containerd.Client, dgst diges
...
-				}
-			}
-			if !containsDefault {
-				return "", nil, nil, nil // do not run analyzer
-			}
+	var paths []string
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		line := strings.TrimSpace(scanner.Text())
+		if line != "" && !strings.HasPrefix(line, "#") {
```
ktock, line 273: The usage of this flag and the file syntax should be documented somewhere such as `docs/ctr-remote.md`

PR #2117, opened 2025-09-04
Using decompression commands to improve the layer decompression speed of gzip-formatted images

escapefreeg:
The current code uses Go’s built-in gzip library to decompress gzip-formatted images. However, it was found that decompression using Go’s gzip library is slow (especially for large images), while using command-line decompression tools can improve the decompression speed. In the below 10 examples, using decompression commands achieved at least 32% speedup. 
image
The containerd community has also discussed and implemented related optimizations, see: https://github.com/containerd/containerd/pull/2640.

Although containerd uses the igzip command for decompression by default (see: https://github.com/containerd/containerd/blob/02eb68472ea11518c419d8866a3a309826bbb3ab/pkg/archive/compression/compression.go#L270 ), above results show that pigz delivers the best performance across most test cases. Therefore, the current priority for invoking decompression commands is pigz > igzip > gzip.

### DIFF
--- a/cmd/ctr-remote/commands/convert.go
+++ b/cmd/ctr-remote/commands/convert.go
@@ -37,6 +37,7 @@ import (
 	esgzexternaltocconvert "github.com/containerd/stargz-snapshotter/nativeconverter/estargz/externaltoc"
 	zstdchunkedconvert "github.com/containerd/stargz-snapshotter/nativeconverter/zstdchunked"
 	"github.com/containerd/stargz-snapshotter/recorder"
+	"github.com/containerd/stargz-snapshotter/util/decompressutil"
 	"github.com/klauspost/compress/zstd"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/urfave/cli/v2"
@@ -86,6 +87,12 @@ When '--all-platforms' is given all images in a manifest list must be available.
 		&cli.BoolFlag{
 			Name:  "estargz-keep-diff-id",
 			Usage: "convert to esgz without changing diffID (cannot be used in conjunction with '--estargz-record-in'. must be specified with '--estargz-external-toc')",
+		},
+		&cli.StringFlag{
+			Name:    "estargz-gzip-helper",
+			Aliases: []string{"GH"},
+			Usage:   "Helper command for decompressing layers compressed with gzip. Options: pigz, igzip, or gzip.",
+			Value:   "", // see also https://github.com/containerd/stargz-snapshotter/pull/2117
 		},
 		// zstd:chunked flags
 		&cli.BoolFlag{
@@ -280,6 +287,13 @@ func getESGZConvertOpts(context *cli.Context) ([]estargz.Option, error) {
 		esgzOpts = append(esgzOpts, estargz.WithPrioritizedFiles(paths))
 		var ignored []string
 		esgzOpts = append(esgzOpts, estargz.WithAllowPrioritizeNotFound(&ignored))
+	}
+	if estargzGzipHelper := context.String("estargz-gzip-helper"); estargzGzipHelper != "" {
+		gzipHelperFunc, err := decompressutil.GetGzipHelperFunc(estargzGzipHelper)
+		if err != nil {
+			return nil, err
+		}
+		esgzOpts = append(esgzOpts, estargz.WithGzipHelperFunc(gzipHelperFunc))
 	}
 	return esgzOpts, nil
 }
--- a/cmd/ctr-remote/commands/optimize.go
+++ b/cmd/ctr-remote/commands/optimize.go
@@ -189,21 +189,22 @@ var OptimizeCommand = &cli.Command{
 		if clicontext.Bool("zstdchunked") {
 			f = zstdchunkedconvert.LayerConvertWithLayerOptsFuncWithCompressionLevel(
 				zstd.EncoderLevelFromZstd(clicontext.Int("zstdchunked-compression-level")), esgzOptsPerLayer)
-		} else if !clicontext.Bool("estargz-external-toc") {
-			f = estargzconvert.LayerConvertWithLayerAndCommonOptsFunc(esgzOptsPerLayer,
-				estargz.WithCompressionLevel(clicontext.Int("estargz-compression-level")),
-				estargz.WithChunkSize(clicontext.Int("estargz-chunk-size")),
-				estargz.WithMinChunkSize(clicontext.Int("estargz-min-chunk-size")))
 		} else {
-			if clicontext.Bool("reuse") {
+			esgzOpts := []estargz.Option{
+				estargz.WithChunkSize(clicontext.Int("estargz-chunk-size")),
+				estargz.WithMinChunkSize(clicontext.Int("estargz-min-chunk-size")),
+			}
+			if !clicontext.Bool("estargz-external-toc") {
+				esgzOpts = append(esgzOpts, estargz.WithCompressionLevel(clicontext.Int("estargz-compression-level")))
+				f = estargzconvert.LayerConvertWithLayerAndCommonOptsFunc(esgzOptsPerLayer, esgzOpts...)
+			} else if clicontext.Bool("reuse") {
 				// We require that the layer conversion is triggerd for each layer
 				// to make sure that "finalize" function has the information of all layers.
 				return fmt.Errorf("\"estargz-external-toc\" can't be used with \"reuse\" flag")
+			} else {
+				f, finalize = esgzexternaltocconvert.LayerConvertWithLayerAndCommonOptsFunc(esgzOptsPerLayer, esgzOpts,
+					clicontext.Int("estargz-compression-level"))
 			}
-			f, finalize = esgzexternaltocconvert.LayerConvertWithLayerAndCommonOptsFunc(esgzOptsPerLayer, []estargz.Option{
-				estargz.WithChunkSize(clicontext.Int("estargz-chunk-size")),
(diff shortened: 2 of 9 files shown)


estargz/decompressutil.go
```diff
@@ -0,0 +1,117 @@
...
+	if err != nil {
+		return path
+	}
+
+	if disable {
+		return ""
+	}
+
+	return path
+}
```
ChengyuZhu6: I recommend to merge `findCmdPathWithLookuper` and `findCmdPath` into one function named `findCmdPath`.

estargz/decompressutil_test.go
```diff
@@ -0,0 +1,168 @@
...
+		})
+	}
+}
+
+func TestRunCmdDecompress(t *testing.T) {
+	refreshCmdPath()
+
+	if gzipPath == "" && pigzPath == "" && igzipPath == "" {
+		t.Skip("Skipping test: gzip, pigz or igzip commands not available in environment")
+	}
```
ChengyuZhu6: Is the test always skipped? I think we should preinstall pigz or igzip in the CI test image. https://github.com/containerd/stargz-snapshotter/blob/a2df5f8fc62ede343e2f115340b7f2cc41142137/script/util/make.sh#L37

ChengyuZhu6 (commented):
Some minor comments.

estargz/decompressutil.go
escapefreeg: Thanks for the review. Since I want to test the `findCmdPath` function in `estargz/decompressutil_test.go` without being affected by the local environment, I introduced the `findCmdPathWithLookuper` function. By providing a custom `mockLookuperFunc`, we can simulate locating a command’s path, making the tests deterministic and environment-independent.

estargz/decompressutil_test.go
escapefreeg: done

estargz/build.go
```diff
@@ -660,6 +660,22 @@ func decompressBlob(org *io.SectionReader, tmp *tempFiles) (*io.SectionReader, e
 	var dR io.Reader
 	if bytes.Equal([]byte{0x1F, 0x8B, 0x08}, src[:3]) {
 		// gzip
+		// Attempt decompression using commands in priority order: pigz > igzip > gzip
+		cmdPath := getDecompressCmdPath()
```
ktock: This library doesn't intend to implicitly execute commands. This should be an optional behaviour and the option should be just a callback function with a type smtg like `func(io.Reader) (io.Reader, error)` with an implementation outside of  the `estargz` module . Also, you can add a flag for `ctr-remote` like `--decompression-helper` for easy testing.

escapefreeg: Thanks for your suggestion. I have made changes according to your advice. Could you please check if there are any other issues?

cmd/ctr-remote/commands/convert.go
```diff
@@ -87,6 +88,12 @@ When '--all-platforms' is given all images in a manifest list must be available.
 			Name:  "estargz-keep-diff-id",
 			Usage: "convert to esgz without changing diffID (cannot be used in conjunction with '--estargz-record-in'. must be specified with '--estargz-external-toc')",
 		},
+		&cli.StringFlag{
+			Name:    "estargz-gzip-helper",
```
wswsmao, line 92: New options have been added, please update the ctr-remote.md

util/decompressutil/gzip.go
```diff
@@ -0,0 +1,141 @@
...
+   limitations under the License.
+*/
+
+package decompressutil
+
+import (
+	"bytes"
+	"compress/gzip"
+	"fmt"
+	"github.com/containerd/stargz-snapshotter/estargz"
```
ktock: Can you fix the linter error?

> File is not properly formatted (goimports)

util/decompressutil/gzip.go
```diff
@@ -0,0 +1,141 @@
...
+	})
+}
+
+func getCmdGzipHelperFunc(cmdPath string) estargz.GzipHelperFunc {
+	return func(in io.Reader) (io.ReadCloser, error) {
+		cmd := exec.Command(cmdPath, "-d", "-c")
+
+		readCloser, writer := io.Pipe()
+		cmd.Stdin = in
+		cmd.Stdout = writer
```
ktock, line 51: Can we just use StdoutPipe? https://pkg.go.dev/os/exec#Cmd.StdoutPipe

cmd/ctr-remote/commands/convert.go
escapefreeg: done

util/decompressutil/gzip.go
escapefreeg: done

util/decompressutil/gzip.go
escapefreeg: I have tried this approach, but when an error occurs in the goroutine that runs `cmd.Wait`, the `io.ReadCloser` returned by `cmd.StdoutPipe` cannot notify the decompression reader side that the data has errored, which causes the program to behave abnormally. If don’t use a pipe, I might need to implement a custom ReadCloser wrapper or use context to handle the write and error logic, which could increase code complexity.

In addition, I found that `cmd.StdoutPipe` internally also uses a pipe, as shown here: https://cs.opensource.google/go/go/+/refs/tags/go1.25.1:src/os/exec/exec.go;drc=ebee011a54f9310099d02a7e7731330539db16cf;l=1074.

Therefore, I have not made changes to the related code for now. If there is a better implementation, please let me know.

cmd/ctr-remote/commands/optimize.go
```diff
@@ -189,21 +196,35 @@ var OptimizeCommand = &cli.Command{
...
+			if estargzGzipHelper := clicontext.String("estargz-gzip-helper"); estargzGzipHelper != "" {
+				gzipHelperFunc, err := decompressutil.GetGzipHelperFunc(estargzGzipHelper)
+				if err != nil {
+					return err
+				}
+				esgzBaseOpts = append(esgzBaseOpts, estargz.WithGzipHelperFunc(gzipHelperFunc))
+			}
+			if !clicontext.Bool("estargz-external-toc") {
+				// copy esgzBaseOpts to esgzOpts to avoid data race
+				esgzOpts := make([]estargz.Option, len(esgzBaseOpts)+1)
```
ktock: Coud you elaborate this? Is this accessed from multiple goroutines?

escapefreeg: The callback function returned by `estargzconvert.LayerConvertWithLayerAndCommonOptsFunc` captures the slice `esgzOpts`. When different layers in an image invoke the conversion function `f` concurrently, the operation inside `f`: `LayerConvertFunc(append(commonOpts, opts[desc.Digest]...)...)(ctx, cs, desc)` will append to the captured `esgzOpts` simultaneously, which can lead to a data race.

If we don’t make a copy, the check for the `estargz-gzip-helper` flag may appends option to `esgzBaseOpts`, causing its len and cap  properties to become unequal. when`LayerConvertFunc(append(commonOpts, opts[desc.Digest]...)...)(ctx, cs, desc)`executes, concurrent layers may append to the same underlying slice, resulting in a data race.

If we perform a copy, the new `esgzOpts` always has its len property equal to its cap property. In that case, the append in `LayerConvertFunc(append(commonOpts, opts[desc.Digest]...)...)(ctx, cs, desc)` will always allocate a new slice due to insufficient capacity, so each layer gets its own private slice and no data race occurs.

The use of copy in `esgzexternaltocconvert.LayerConvertWithLayerAndCommonOptsFunc` follows the same rationale.

The original code has a similar phenomenon. For example, `estargzconvert.LayerConvertWithLayerAndCommonOptsFunc` constructs a temporary `estargz.Option` slice on the fly, ensuring that its len property is equal to its cap property. However, this is a subtle issue that could cause problems in future development. When I submitted the commits, I didn’t fully consider this and went with the current approach. A better solution would be to copy the `estargz.Option` slice inside both `estargzconvert.LayerConvertWithLayerAndCommonOptsFunc` and `esgzexternaltocconvert.LayerConvertWithLayerAndCommonOptsFunc` to completely eliminate the data race. I’ll submit a PR to address this.

ktock (approved):
Thanks

docs/ctr-remote.md
```diff
@@ -58,6 +58,14 @@ ctr-remote image push --plain-http registry2:5000/golang:1.15.3-esgz
 When you run `ctr-remote image optimize`, this runs the source image (`ghcr.io/stargz-containers/golang:1.15.3-buster-org`) as a container and profiles all file accesses during the execution.
 Then these accessed files are marked as "prioritized" files and will be prefetched on runtime.
 
+You can specify the `--estargz-gzip-helper` flag to specify a command-line helper tool for gzip decompression during optimization. Using command-line tools can speed up the decompression of the original image, thereby accelerating the overall optimization process. Even using the gzip command corresponding to the Go gzip library can achieve approximately 32% speed improvement. For more details, see: [Using decompression commands to improve the layer decompression speed of gzip-formatted images](https://github.com/containerd/stargz-snapshotter/pull/2117). Currently, `--estargz-gzip-helper` only supports `pigz`, `igzip`, and `gzip`.
```
AkihiroSuda, line 61: Does this still work with lazy pulling?

escapefreeg: Yes, the `--estargz-gzip-helper` is only used to decompress tar.gz-encoded traditional overlayfs images.  It does not change the subsequent image conversion/optimization process, and therefore does not affect the lazy pulling of estargz format images. I also converted/optimized the same image using the code before and after the modification. The manifest and overall hash of the converted/optimized image remained identical in both cases.

Issue #2126, opened 2025-09-12
Normal snapshots lost after stargz snapshotter restart, causing container startup failures


ChengyuZhu6:
When stargz-grpc is stopped and restarted, normal (non-remote) snapshots are lost, causing containers to fail to start with errors like:
```
FATA[0000] failed to stat parent: stat /var/lib/containerd-stargz-grpc/snapshotter/snapshots/1/fs: no such file or directory
```

## Steps to Reproduce

1. **Start a non-stargz image using stargz snapshotter:**
   ```bash
   nerdctl run --snapshotter stargz -it ubuntu:20.04 /bin/bash
   ```
   - Image pulls successfully
   - Container starts and works normally
   - Snapshots are created: `/var/lib/containerd-stargz-grpc/snapshotter/snapshots/1` and `3`

2. **Exit the container and stop stargz-grpc:**
   ```bash
   # Exit container
   exit
   # Stop snapshotter with Ctrl+C
   ```

3. **Restart stargz-grpc and try to run the same image:**
   ```bash
   nerdctl run --snapshotter stargz -it ubuntu:20.04 /bin/bash
   ```
   - **Expected**: Container should start successfully (reuse existing snapshots)
   - **Actual**: Fails with "no such file or directory" error

The issue occurs because:

1. **During shutdown**: The `Close()` method calls `cleanup(ctx, cleanupCommitted=true)` which deletes ALL snapshot directories, including normal (non-remote) snapshots
2. **During startup**: The `restoreRemoteSnapshot()` method only restores remote snapshots (those with `remoteLabel`), but normal snapshots are not restored
3. **Result**: Normal snapshot directories are deleted but not recreated, causing the "no such file or directory" error when trying to use them as parent snapshots

## Expected Behavior

- Normal snapshots should be preserved across stargz-grpc restarts
- The same image should be able to reuse existing snapshots after restart
- No "no such file or directory" errors should occur

PR #2127, opened 2025-09-12
snapshot: preserve normal snapshots during cleanup

ChengyuZhu6:
This commit modifies getCleanupDirectories() to:
- Only clean up remote snapshots (those with remoteLabel) during Close()
- Preserve normal snapshot directories and their content
- Ensure normal snapshots remain functional after restart

Fixes: #2126

### DIFF
--- a/snapshot/snapshot.go
+++ b/snapshot/snapshot.go
@@ -455,10 +455,32 @@ func (o *snapshotter) getCleanupDirectories(ctx context.Context, t storage.Trans
 		return nil, err
 	}
 
+	remoteSnapshotNames := make(map[string]struct{})
+	{
+		keyToID := make(map[string]string, len(ids))
+		for id, key := range ids {
+			keyToID[key] = id
+		}
+		if err := storage.WalkInfo(ctx, func(ctx context.Context, info snapshots.Info) error {
+			if _, ok := info.Labels[remoteLabel]; ok {
+				if id, exists := keyToID[info.Name]; exists {
+					remoteSnapshotNames[id] = struct{}{}
+				}
+			}
+			return nil
+		}); err != nil {
+			return nil, err
+		}
+	}
+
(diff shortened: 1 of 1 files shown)


snapshot/snapshot.go
```diff
@@ -455,12 +455,31 @@ func (o *snapshotter) getCleanupDirectories(ctx context.Context, t storage.Trans
...
+	if cleanupCommitted {
+		keyToID := make(map[string]string, len(ids))
+		for id, key := range ids {
+			keyToID[key] = id
+		}
+
+		if err := storage.WalkInfo(ctx, func(ctx context.Context, info snapshots.Info) error {
+			if _, ok := info.Labels[remoteLabel]; !ok {
+				if id, exists := keyToID[info.Name]; exists {
+					delete(ids, id)
```
ktock: Instead of giving two different meanings to `ids` depending on `cleanupCommited` value, can you just add a `remoteSnapshotNames := make(map[string]struct{})` which caches all remote snapshot names and do something like the following in the loop at L459?

```go
for _, d := range dirs {
	if !cleanupCommitted {
		if _, ok := ids[d]; ok {
			continue
		}
	} else {
		if _, ok := remoteSnapshotNames[d]; !ok {
			continue // do not delete local snapshots
		}
	}

	cleanup = append(cleanup, filepath.Join(snapshotDir, d))
}
```

ChengyuZhu6: Sure. I'll make the change.

ChengyuZhu6: Done.

ktock:
@ChengyuZhu6 Could you rebase the patch to pass CI?

ChengyuZhu6:
> @ChengyuZhu6 Could you rebase the patch to pass CI?

Sure.

ktock (approved):
Thanks

Issue #2128, opened 2025-09-12
Prefetch miss with GPU containers due to missing pre-container file access tracking


wswsmao:
## Problem Description

When running containers with the `--gpus all` option, optimized images created by the `optimize` command show poor prefetch performance with significant prefetch misses.

## Root Cause Analysis

The issue stems from a timing gap in file access tracking:

1. **Pre-container OCI Hook Execution**: NVIDIA runtime injects dynamic libraries and executes `ldconfig` via OCI hooks before the container actually starts
2. **Early File Access**: The `ldconfig` operation accesses various dynamic libraries in the image during this pre-container phase
3. **Missing Tracking**: The current `optimize` process only starts fanotify monitoring after the container is running, missing the file access patterns from steps 1-2
4. **Runtime Prefetch Misses**: When the container runs, the same OCI hook behavior repeats, causing FUSE to trigger on-demand loading for files that should have been prefetched

## Proposed Solution

Implement a pre-container monitor that:
- Starts tracking file access before container creation
- Use FAN_MARK_FILESYSTEM to run fanotify on the host. (Testing has shown that FAN_MARK_MOUNT is unable to track a container's rootfs).
- Captures access patterns from OCI hooks and pre-container operations

PR #2129, opened 2025-09-12
analyzer: add pre-container monitor to capture early file access

wswsmao:
Fixes: https://github.com/containerd/stargz-snapshotter/issues/2128

### DIFF
--- a/analyzer/analyzer.go
+++ b/analyzer/analyzer.go
@@ -41,6 +41,7 @@ import (
 	"github.com/containerd/log"
 	"github.com/containerd/platforms"
 	"github.com/containerd/stargz-snapshotter/analyzer/fanotify"
+	"github.com/containerd/stargz-snapshotter/analyzer/fanotify/service"
 	"github.com/containerd/stargz-snapshotter/analyzer/recorder"
 	"github.com/opencontainers/go-digest"
 	"github.com/opencontainers/image-spec/identity"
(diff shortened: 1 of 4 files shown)


analyzer/analyzer.go
```diff
@@ -195,6 +216,21 @@ func Analyze(ctx context.Context, client *containerd.Client, ref string, opts ..
 		return "", err
 	}
 	defer rc.Close()
+
+	if preMonitorEnabled {
+		prePaths := preMonitor.GetPaths()
+		for _, path := range prePaths {
+			cleanPath := path
+			if strings.HasPrefix(path, target) {
```
ktock, line 224: Is it possible that this becomes false? If so, that path should not be recorded and just be skipped.

wswsmao: I investigated this on `nvcr.io/nvidia/tritonserver:25.07-py3` and found that during the pre-monitor phase fanotify can report the same file in two forms: with and without the `target` mount prefix. For example:

- Without prefix:
  - `/usr/lib/x86_64-linux-gnu/libubsan.so.1.0.0`
  - `/usr/lib/x86_64-linux-gnu/liblber.so.2.0.200`
  - `/opt/hpcx/ompi/lib/libpmix.so.2.2.35`
- With prefix:
  - `/tmp/target296403442/usr/bin/nvidia-persistenced`
  - `/tmp/target296403442/usr/lib/firmware/nvidia/535.261.03/gsp_ga10x.bin`

After the container starts, these files are present in the container filesystem as expected:
```
# inside the running container
ls /usr/lib/x86_64-linux-gnu/libubsan.so.1.0.0 \
   /usr/lib/x86_64-linux-gnu/liblber.so.2.0.200 \
   /opt/hpcx/ompi/lib/libpmix.so.2.2.35
# outputs: all three paths exist

ls /usr/bin/nvidia-persistenced \
   /usr/lib/firmware/nvidia/535.261.03/gsp_ga10x.bin
# outputs: both paths exist
```

So `strings.HasPrefix(path, target)` can legitimately be false for valid accesses under the rootfs. If we skip those, we will miss real file accesses. 
The safer approach is to normalize: when the prefix is present, trim it; otherwise record the path as-is.  Even if there is a non-existent file, `recorder.Record` will skip it here.

ktock (approved):
Thanks

PR #2135, opened 2025-09-25
fusemanager: Log by default

mmonaco:
## fusemanager: Drop --debug flag (not used)
There's `--log-level` instead.

## fusemanager: Hookup normal stdout+stderr by default
Dropping logging output by default is surprising. I think switching this as the default is safe to do for existing users.

## fusemanager: Drop manual truncation of logPath
`os.Create()` truncates existing files already, but additionally works
cleanly if the path is a literal `/dev/stdout` or `/dev/stderr` which
would otherwise fail to be removed.

### DIFF
--- a/fusemanager/fusemanager.go
+++ b/fusemanager/fusemanager.go
@@ -117,10 +117,6 @@ func startNew(ctx context.Context, logPath, address, fusestore, logLevel string)
 		cmd.Stdout = os.Stdout
 		cmd.Stderr = os.Stderr
 	} else {
-		err := os.Remove(logPath)
-		if err != nil && !os.IsNotExist(err) {
-			return err
-		}
 		file, err := os.Create(logPath)
 		if err != nil {
 			return err


ktock (approved):
Thanks. I'll merge this after fixing the CI failure (#2136).

mmonaco:
> @mmonaco Could you add "Signed-off-by" line to the commit message? Also, please rebase on the latest main branch to fix the following CI failures:
> 
> ```
> 148.2 vendor/github.com/containerd/stargz-snapshotter/service/keychain/cri/cri.go:54:29: cannot use server (variable of type *instrumentedService) as "k8s.io/cri-api/pkg/apis/runtime/v1".ImageServiceServer value in return statement: *instrumentedService does not implement "k8s.io/cri-api/pkg/apis/runtime/v1".ImageServiceServer (missing method mustEmbedUnimplementedImageServiceServer)
> ```

Done, and thanks :)

ktock (approved):
Thanks

PR #2145, opened 2025-10-13
avoid repeated decompression and further utilize the --GH option to improve conversion speed

escapefreeg:
During the current image conversion process, the converted blob is decompressed in `estargz/build.go` to calculate its hash value. Additionally, the converted blob is decompressed again in `nativeconverter/estargz/estargz.go` (estargz format) and `nativeconverter/zstdchunked/zstdchunked.go` (zstd format) to calculate its size. However, these two decompression operations are not necessary, and decompression is a time-consuming task. Therefore, the two decompressions can be merged into a single operation to improve image conversion speed.

Furthermore, according to #2117 , compared with Go’s built-in gzip decompression library, command-line tools can achieve faster decompression of gzip archives. When the `--estargz-gzip-helper` (`--GH`) option is specified, the gzip helper can be further utilized to speed up decompression of the converted image.

I' conducted tests on the related changes. Results from 10 images show that the improvement reduces conversion time by an average of 26.7%. For more details, see:
image

The calculation method for `Time improvement from applying current optimization on pigz GH` is (`With pigz GH` - `With pigz GH and current optimization`) / `With pigz GH` and the calculation method for `Time improvement from combining pigz GH and current optimization` is (`Without GH` - `With pigz GH and current optimization`) / `Without GH`

Note: Although `ioutils.CountWriter` ensures correct byte counting in concurrent environments, its usage in `nativeconverter/estargz/estargz.go` and `nativeconverter/zstdchunked/zstdchunked.go` does not involve concurrency. Therefore, the size of the converted blob can be directly obtained from the return value of `io.Copy`.

### DIFF
--- a/estargz/build.go
+++ b/estargz/build.go
@@ -35,6 +35,7 @@ import (
 	"runtime"
 	"strings"
 	"sync"
+	"sync/atomic"
 
 	"github.com/containerd/stargz-snapshotter/estargz/errorutil"
 	"github.com/klauspost/compress/zstd"
@@ -145,8 +146,10 @@ func WithGzipHelperFunc(gzipHelperFunc GzipHelperFunc) Option {
 // Blob is an eStargz blob.
 type Blob struct {
 	io.ReadCloser
-	diffID    digest.Digester
-	tocDigest digest.Digest
+	diffID           digest.Digester
+	tocDigest        digest.Digest
+	readCompleted    *atomic.Bool
+	uncompressedSize *atomic.Int64
 }
 
 // DiffID returns the digest of uncompressed blob.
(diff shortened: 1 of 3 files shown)


estargz/build.go
```diff
@@ -267,26 +273,37 @@ func Build(tarBlob *io.SectionReader, opt ...Option) (_ *Blob, rErr error) {
...
 		}
-		defer r.Close()
-		if _, err := io.Copy(diffID.Hash(), r); err != nil {
+		defer decompressR.Close()
+		if uncompressedSize, err = io.Copy(diffID.Hash(), decompressR); err != nil {
 			pw.CloseWithError(err)
 			return
 		}
+		uncompressedSizeChan <- uncompressedSize
+		close(uncompressedSizeChan)
```
ktock: Does this make `UncompressedSize` return a zero value when it called more than once? And I guess this coroutine leaks with blocking at L295 if `UncompressedSize` isn't called. 
Instead of using channel, I think it can just use an int64 pointer + atomic. `UncompressedSize` can have a comment that the value is valid only after the full read.

escapefreeg: Thanks for the suggestion. I've updated it based on your advice.

PR #2148, opened 2025-10-15
ctr-remote: add wildcard support for prefetch-list option

wswsmao:
https://github.com/containerd/stargz-snapshotter/pull/2113  introduced prefetch-list, but it only supports specific filenames. This commit adds wildcard support for the prefetch-list option, allowing users to flexibly select files to prefetch using wildcards.


analyzer/recorder/recorder.go
```diff
@@ -173,6 +176,46 @@ func (r *ImageRecorder) Record(name string) error {
...
+			return matchedCount, err
+		}
+		if !matched {
+			continue
+		}
+		if err := r.Record(name); err != nil {
+			continue
+		}
+		matchedCount++
+	}
```
ktock, line 201: Could you move this to the optimize.go so that the new dependency isn't added to the root gomod?

wswsmao: done

ktock (approved):
Thanks

PR #2153, opened 2025-10-21
ctr-remote: improve sortEntries performance from O(k·n) to O(n)+O(k)

wswsmao:
### Problem
Testing revealed that `sortEntries` execution time increases significantly with larger `prioritized` input lists. The original implementation calls `in.remove(name)` for each prioritized entry, which rebuilds the entire stream (O(n) operation). When processing k prioritized files, this results in O(k·n) time complexity, causing severe performance degradation with large priority lists.

### Solution
Instead of removing entries one by one, the optimized version:
1. Uses `addRec` to collect prioritized entries and mark them in a `picked` set
2. Performs a single O(n) filtering pass via `dumpExcept` to exclude marked entries
3. Concatenates prioritized and remaining entries

### DIFF
--- a/estargz/build.go
+++ b/estargz/build.go
@@ -410,8 +410,9 @@ func sortEntries(in io.ReaderAt, prioritized []string, missedPrioritized *[]stri
 
 	// Sort the tar file respecting to the prioritized files list.
 	sorted := &tarFile{}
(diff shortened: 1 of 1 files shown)


ktock (approved):
Thanks

Issue #2178, opened 2025-11-21
Implement Hot Reloading for Configuration


wswsmao:
#### **Background**

Many configuration options, particularly those found in `fs/config/config.go`, can be tuned to optimize the user experience and performance of `stargz-snapshotter`. Currently, applying any configuration change requires a full restart of the snapshotter. This process can be disruptive and costly for users, especially in production environments where running containers depend on the snapshotter and must also be restarted.

#### **Proposal**

To address this limitation, we can introduce a configuration hot reloading mechanism. This would allow users to apply configuration changes dynamically without interrupting service.

The implementation could follow these steps:

1.  **Monitor Configuration File Changes:** Use a library like `fsnotify` to watch the configuration file for modifications.
2.  **Propagate Configuration Updates:** Add methods to relevant components—from the top-level snapshotter down to the layer management logic—to support dynamic configuration updates.
3.  **Chain of Updates:** When a configuration change is detected, trigger a chained update sequence that propagates the new settings throughout the system.

#### **Benefits**

With this feature, newly pulled images would immediately benefit from the updated configuration, while existing operations remain unaffected. This would significantly improve the flexibility and manageability of `stargz-snapshotter` in dynamic environments.

PR #2179, opened 2025-11-21
add hot reloading for fs config

wswsmao:
Fixes: https://github.com/containerd/stargz-snapshotter/issues/2178

This PR implements hot reloading, but only for fs configurations. My rationale is to reserve this feature for settings that are truly optional improvements.

The fs settings fit this model perfectly: the system is fully functional out-of-the-box, but users can tweak them for better results. Because the "best" value can be subjective or workload-dependent, users may want to adjust them frequently. Hot reloading makes this process of discovery and tuning seamless.

For higher-level settings that control fundamental behavior, hot reloading offers little benefit and introduces potential risk. A restart is the more appropriate mechanism for enabling or altering core features.

### DIFF
--- a/cmd/containerd-stargz-grpc/cfgwatcher.go
+++ b/cmd/containerd-stargz-grpc/cfgwatcher.go
@@ -1,0 +1,152 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package main
+
+import (
+	"context"
+	"path/filepath"
+	"reflect"
+	"sync"
+	"time"
+
+	"github.com/containerd/containerd/v2/core/snapshots"
+	"github.com/containerd/log"
+	fsconfig "github.com/containerd/stargz-snapshotter/fs/config"
+	"github.com/fsnotify/fsnotify"
+	"github.com/pelletier/go-toml"
+)
+
+// WatchConfig monitors the specified configuration file for changes.
+// It triggers the config reload when a change is detected.
+func WatchConfig(
+	ctx context.Context,
+	filePath string,
+	rs snapshots.Snapshotter,
+	initialConfig *fsconfig.Config,
+) error {
+	absFilePath, err := filepath.Abs(filePath)
+	if err != nil {
+		return err
+	}
+
+	watchDir := filepath.Dir(absFilePath)
+
+	watcher, err := fsnotify.NewWatcher()
+	if err != nil {
+		return err
+	}
+
+	if err := watcher.Add(watchDir); err != nil {
+		watcher.Close()
+		return err
+	}
+
+	log.G(ctx).Infof("started monitoring config file: %s", absFilePath)
+
+	cw := &configWatcher{
+		lastConfig: initialConfig,
+	}
+
+	go func() {
+		defer watcher.Close()
+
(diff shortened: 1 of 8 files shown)


wswsmao:
@ktock

cmd/containerd-stargz-grpc/cfgwatcher.go
```diff
@@ -0,0 +1,182 @@
...
+		return
+	}
+
+	newFsConfig := newConfig.Config.Config
+
+	w.mu.Lock()
+	defer w.mu.Unlock()
+
+	if w.lastConfig != nil {
+		revertBlacklistedChanges(ctx, w.lastConfig, &newFsConfig)
```
ktock: It looks like each package manages its own whitelist in UpdateConfig. Why do we need an additional blacklist here?

docs/overview.md
```diff
@@ -260,6 +260,13 @@ insecure = true
 
 The config file can be passed to stargz snapshotter using `containerd-stargz-grpc`'s `--config` option.
 
+## Configuration hot reload
```
ktock, line 263: The benefit of this feature compared to the FUSE manager (which already supports safe restarts) should be documented.

cmd/containerd-stargz-grpc/cfgwatcher.go
wswsmao: done, remove it

docs/overview.md
wswsmao: done

wswsmao:
All reviews done @ktock

cmd/containerd-stargz-grpc/cfgwatcher.go
```diff
@@ -0,0 +1,152 @@
...
+			case event, ok := <-watcher.Events:
+				if !ok {
+					return
+				}
+
+				if event.Name == absFilePath {
+					// Trigger on Write, Create, Rename, or Chmod events
+					// such as vim, nano, etc.
+					if event.Has(fsnotify.Write) || event.Has(fsnotify.Create) ||
+						event.Has(fsnotify.Rename) || event.Has(fsnotify.Chmod) {
```
ktock, line 83: > fsnotify.Rename

If the file is renamed, the next change won't be detected, will it? Is this an expected behaviour?

cmd/containerd-stargz-grpc/cfgwatcher.go
```diff
@@ -0,0 +1,152 @@
...
+	"github.com/containerd/containerd/v2/core/snapshots"
+	"github.com/containerd/log"
+	fsconfig "github.com/containerd/stargz-snapshotter/fs/config"
+	"github.com/fsnotify/fsnotify"
+	"github.com/pelletier/go-toml"
+)
+
+// WatchConfig monitors the specified configuration file for changes.
+// It triggers the config reload when a change is detected.
+func WatchConfig(
```
ktock, line 35: You don't need to export this symbol.

cmd/containerd-stargz-grpc/cfgwatcher.go
```diff
@@ -0,0 +1,152 @@
...
+	cw := &configWatcher{
+		lastConfig: initialConfig,
+	}
+
+	go func() {
+		defer watcher.Close()
+
+		var (
+			debounceTimer *time.Timer
+			mu            sync.Mutex
```
ktock, line 69: Why is `debounceTimer` protected by the mutex? Is this accessed from multiple goroutines? I cound't find new goroutine creation from inside of this function.

cmd/containerd-stargz-grpc/cfgwatcher.go
```diff
@@ -0,0 +1,152 @@
...
+	initialConfig *fsconfig.Config,
+) error {
+	absFilePath, err := filepath.Abs(filePath)
+	if err != nil {
+		return err
+	}
+
+	watchDir := filepath.Dir(absFilePath)
+
+	watcher, err := fsnotify.NewWatcher()
```
ktock, line 48: Does this work in the [rootless settings](https://github.com/containerd/stargz-snapshotter/blob/76ac7199ee68980e251393bc5a231e04b5779756/docs/rootless.md)?

PR #2194, opened 2025-12-09
k3s: fix CI failure by freeing up more disk space

ktock:
The test for the k3s main is broken in recent days. Fix it by locking the version to the stable one : https://github.com/containerd/stargz-snapshotter/actions/runs/19996164485/job/57343963732#step:7:5003


ktock:
Still failing

> Deadline exeeded to wait for pod creation

PR #2205, opened 2025-12-25
Add log file access information on first access

wswsmao:
This PR introduces a file access log in debug mode.
By using a sync.Map to track accessed node IDs, it ensures that metadata for each file—including its path, size, mode, and layer digest—is logged only once when it is first opened or its symlink is read. 
This feature is useful for analyzing container image usage patterns and identifying which files are actually being accessed during runtime.

### DIFF
--- a/fs/layer/node.go
+++ b/fs/layer/node.go
@@ -30,9 +30,11 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path"
 	"sort"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"syscall"
 	"time"
 
@@ -76,7 +78,7 @@ var opaqueXattrs = map[OverlayOpaqueType][]string{
 	OverlayOpaqueUser:    {"user.overlay.opaque"},
 }
 
-func newNode(layerDgst digest.Digest, r reader.Reader, blob remote.Blob, baseInode uint32, opaque OverlayOpaqueType, pth passThroughConfig) (fusefs.InodeEmbedder, error) {
+func newNode(layerDgst digest.Digest, r reader.Reader, blob remote.Blob, baseInode uint32, opaque OverlayOpaqueType, pth passThroughConfig, logFileAccess bool) (fusefs.InodeEmbedder, error) {
 	rootID := r.Metadata().RootID()
 	rootAttr, err := r.Metadata().GetAttr(rootID)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


fs/layer/node.go
```diff
@@ -104,19 +105,35 @@ func newNode(layerDgst digest.Digest, r reader.Reader, blob remote.Blob, baseIno
...
-	opaqueXattrs []string
-	passThrough  passThroughConfig
+	r             reader.Reader
+	s             *state
+	layerDigest   digest.Digest
+	baseInode     uint32
+	rootID        uint32
+	opaqueXattrs  []string
+	passThrough   passThroughConfig
+	accessedPaths sync.Map
```
ktock: The size of this data continuously increase during runtime as long as the filesystem is mounted, right? If so, can this feature be optional? And what is the reason we avoid duplication here?

wswsmao: Deduplication is essential to prevent log flooding and keep the debug output clear.
I've replaced sync.Map with a bitset to minimize memory usage (1 bit/file).

fs/layer/node.go
```diff
@@ -111,12 +112,40 @@ type fs struct {
...
 
+func (fs *fs) logAccessOnce(ctx context.Context, n *node) {
+	if fs == nil || n == nil {
+		return
+	}
+	idx := int(n.id / 64)
+	bit := uint(n.id % 64)
+	fs.accessedMu.Lock()
+	if idx >= len(fs.accessedBits) {
+		newBits := make([]uint64, idx+1)
```
ktock: There is no guarantee that the id is assigned incrementally from zero and any number is allowed as long as it fits uint32. If the metadata pkg uses the maximum value (4294967295), this immediately allocates hundreds of MBs of the slice, which is still problematic for the memory efficiency.

wswsmao: I have moved the deduplication flag into the node struct using atomic CAS, which eliminates large global allocations and allows the memory to be automatically reclaimed alongside the inode's lifecycle.

fs/layer/node.go
```diff
@@ -357,6 +375,8 @@ func (n *node) Open(ctx context.Context, flags uint32) (fh fusefs.FileHandle, fu
 		return nil, 0, syscall.EIO
 	}
 
+	n.logAccessOnce(ctx)
```
ktock, line 380: Can you enable this function only when a config filed like `log_file_access` is set to true? Debug level logs are  usually used for debugging containerd-stargz-grpc's internal behaviour (e.g. for fixing bugs) but the purpose of this log is different (i.e. for analyzing container image usage patterns) and can potentially increase the amount logs significantly.

fs/layer/node.go
```diff
@@ -437,6 +457,7 @@ func (n *node) Listxattr(ctx context.Context, dest []byte) (uint32, syscall.Errn
 var _ = (fusefs.NodeReadlinker)((*node)(nil))
 
 func (n *node) Readlink(ctx context.Context) ([]byte, syscall.Errno) {
+	n.logAccessOnce(ctx)
```
ktock, line 462: ditto.

fs/layer/node.go
wswsmao: It looks like we need to add a new parameter to newNode to pass this configuration.
However, the parameter list for newNode is already quite long. I'm considering refactoring it via https://github.com/containerd/stargz-snapshotter/pull/2223 first, and then implementing this change based on the refactored code.

wswsmao: > Can you enable this function only when a config filed like `log_file_access` is set to true? Debug level logs are usually used for debugging containerd-stargz-grpc's internal behaviour (e.g. for fixing bugs) but the purpose of this log is different (i.e. for analyzing container image usage patterns) and can potentially increase the amount logs significantly.

done

fs/layer/node.go
wswsmao: done

PR #2206, opened 2025-12-26
fix --reuse being ignored when --no-optimize is set

wswsmao:
Description:
This PR fixes a bug where the optimize command fails to reuse existing eStargz layers if the --no-optimize flag is set.

Changes:
- Refactored analyze to continue to the reuse check logic even when --no-optimize is provided.
- Updated buildLayerOptsFromRecord to handle empty record inputs, allowing it to perform layer reuse validation independently of workload analysis.

This significantly improves conversion efficiency in batch processing scenarios where users want to skip already converted layers without running the analyzer.


ktock:
The patch looks good to me, please rebase to pass CI.

wswsmao:
> The patch looks good to me, please rebase to pass CI.

done

PR #2216, opened 2026-01-07
allow asynchronous prefetch via configurable threshold

wswsmao:
Introduce `PrefetchAsyncSize` to enable non-blocking prefetching when the target size is large.

When the prefetch size of a layer exceeds this threshold, the prefetch waiter is notified immediately. This allows the container to start and initialize without waiting for the entire prefetch process to complete, while the data fetching and decompression continue in the background.

This is particularly beneficial for AI/ML workloads where container images often contain massive model files (e.g., >10GB). By using this feature, the AI framework can start running and loading its environment immediately, while the large model layers are being prefetched in the background, significantly reducing the overall startup latency.


ktock (approved):
Thanks

PR #2223, opened 2026-01-12
refactor newNode to simplify function header

wswsmao:


fs/layer/node.go
```diff
@@ -76,25 +76,26 @@ var opaqueXattrs = map[OverlayOpaqueType][]string{
...
-func newNode(layerDgst digest.Digest, r reader.Reader, blob remote.Blob, baseInode uint32, opaque OverlayOpaqueType, pth passThroughConfig) (fusefs.InodeEmbedder, error) {
+func newNode(l *layer, baseInode uint32) (fusefs.InodeEmbedder, error) {
+	r := l.r
 	rootID := r.Metadata().RootID()
 	rootAttr, err := r.Metadata().GetAttr(rootID)
 	if err != nil {
 		return nil, err
 	}
-	opq, ok := opaqueXattrs[opaque]
+	opq, ok := opaqueXattrs[l.resolver.overlayOpaqueType]
```
ktock, line 86: I think 6-7 params are still fine as this is a private function. `layer` object contains some mutex-protected members so I think it's better to keep newNode agnostic about the internal structure of `layer` to avoid accidentally accessing to those members.

PR #2341, opened 2026-06-12
estargz: add --estargz-parallelism to control build parallelism

simonepri:
eStargz builds fan out across runtime.GOMAXPROCS(0) workers, and the worker count fixes the chunk boundaries. The same image built on machines with different core counts therefore produces different layer digests.

Add `WithParallelism` / `--estargz-parallelism` to pin the worker count, so a build is reproducible regardless of the host's CPU count. `0` (default) keeps GOMAXPROCS; `1` forces a sequential build. It has no effect with `--estargz-min-chunk-size`, which still builds with a single worker.

Note: #2340 parallelizes MinChunkSize builds and touches the same worker-count code in `Build`; whichever lands second will be rebased on the other.

### DIFF
--- a/estargz/build_test.go
+++ b/estargz/build_test.go
@@ -29,6 +29,7 @@ import (
 	"fmt"
 	"io"
 	"reflect"
+	"runtime"
 	"testing"
 )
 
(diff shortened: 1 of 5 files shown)


simonepri:
The two remaining check failures appear unrelated to this change (which is behavior-neutral with default flags):

- **K3S**: also failing on `main` — the last two Nightly runs ([Jun 11](https://github.com/containerd/stargz-snapshotter/actions/runs/27315633431), [Jun 12](https://github.com/containerd/stargz-snapshotter/actions/runs/27386573660)) fail on exactly this job.
- **Integration (false, db, false, true, false)**: the snapshotter daemon never created its socket and produced no output; the same job passed on the two previous runs of this PR with identical content.

Happy to rebase/retrigger once K3S is green on main.

estargz/build_test.go
```diff
@@ -672,3 +673,92 @@ func TestCountReader(t *testing.T) {
...
+
+func incompressibleContent(seed uint64, n int) string {
+	b := make([]byte, n)
+	x := seed*2862933555777941757 + 3037000493
+	for i := range b {
+		x = x*6364136223846793005 + 1442695040888963407
+		b[i] = byte(33 + (x>>33)%94) // printable ASCII
+	}
+	return string(b)
+}
```
ktock: Could you just use `randomContents` from testutil.go?

simonepri: done

simonepri:
@ktock friendly ping :)

simonepri:
@AkihiroSuda any chance to get this merged?

ktock (approved):
Thanks. Please rebase the PR to pass CI.

simonepri:
@ktock done

ktock (approved):

PR #2356, opened 2026-07-22
fusemanager: Add a dedicated metrics endpoint

mironovgh:
In FUSE manager mode, the manager runs in a separate process from the gRPC server. The existing `metrics_address` setting only starts a metrics server in the gRPC server process, so it does not expose metrics produced by the FUSE manager. This makes it difficult to observe production behaviour of the snapshotter.

This change adds a new optional `fusemanager.metrics_address` setting to configure a separate metrics listener for the FUSE manager.

---

Test procedure:

1. Enable the following fuse manager configuration:
```
[fuse_manager]
  enable = true
  metrics_address = "127.0.0.1:6783"
```

2. Restart both gRPC server *and* FUSE manager if it is running
3. Confirm that fuse manager exposes `/metrics` endpoint:
```
$ curl 127.0.0.1:6783/metrics
...
stargz_fs_operation_duration_milliseconds_bucket{layer="sha256:XXX",operation_type="stargz_toc_json_deserialize",le="256"} 1
stargz_fs_operation_duration_milliseconds_bucket{layer="sha256:XXX",operation_type="stargz_toc_json_deserialize",le="512"} 1
...
```

### DIFF
--- a/fusemanager/fusemanager.go
+++ b/fusemanager/fusemanager.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	golog "log"
 	"net"
+	"net/http"
 	"os"
 	"os/exec"
 	"os/signal"
@@ -29,6 +30,7 @@ import (
 	"syscall"
 
 	"github.com/containerd/log"
+	metrics "github.com/docker/go-metrics"
 	"github.com/sirupsen/logrus"
 	"golang.org/x/sys/unix"
 	"google.golang.org/grpc"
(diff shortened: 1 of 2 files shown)


ktock:
Sorry for the slow review. LGTM. 
Please rebase the PR on the latest main to pass the CI.

fusemanager/fusemanager.go
```diff
@@ -189,6 +196,23 @@ func runFuseManager(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("failed to listen socket: %w", err)
 	}
+	defer l.Close()
```
mironovgh, line 199: ensure to cleanup manager's unix socket/file in case metric listener initialization fails and we short-circuit before gRPC server has a chance to start

mironovgh:
@ktock no worries, rebased — looks like there's a data race in ctr-remote 😬.

ktock (approved):
Thanks

PR #2359, opened 2026-08-03
snapshot: restore remote mounts on demand after restart

wxx213:
Add lazy_restore_on_restart to skip remote resolution and FUSE mount restoration during snapshotter startup.

When an unregistered remote layer is first used, restore its mount from the snapshot labels. Coalesce concurrent requests for the same mountpoint and support both in-process and fuse-manager modes.

This allows the snapshotter to start while the registry is unavailable.

### DIFF
--- a/cmd/containerd-stargz-grpc/main.go
+++ b/cmd/containerd-stargz-grpc/main.go
@@ -190,6 +190,9 @@ func main() {
 			log.G(ctx).WithError(err).Fatalf("failed to configure fusemanager")
 		}
 		flags := []snbase.Opt{snbase.AsynchronousRemove}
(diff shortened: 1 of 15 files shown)


ktock:
> This allows the snapshotter to start while the registry is unavailable.

Can you elaborate this? Why is the registry unavailable during the snapshotter startup?

wxx213:
> elaborate

We deployed our image cached registry in cluster by k8s service，most of our application images use it but except like cilium pods. 

On the host reboot，stargz snapshotter cannot connect the registry by service ip to recover remote snapshotters when cilium pod is not ready. So we need the lazy snapshotter recover feature like this.

PR #2361, opened 2026-08-06
Build the k3s node image with make image

kzys:
k3s removed its local-* Makefile targets in k3s-io/k3s#14204, so `make local-image` no longer exists on main and both k3s jobs fail with "No rule to make target 'local-image'".

Drop SKIP_AIRGAP and SKIP_VALIDATE along with it. `image` only runs `binary` and scripts/package-image, neither of which reads them; validation is a separate Dockerfile stage now.


ktock (approved):
Thanks

PR #2363, opened 2026-08-09
fs: add ns query parameter to mirror requests per OCI distribution spec

simonepri:
The `ns` query parameter on pull requests to a mirror identifies the upstream registry the request is for. It is part of the OCI distribution spec ([Registry Proxying](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#registry-proxying)), and containerd adds it to every request it sends to a mirror per its [registry host namespace convention](https://github.com/containerd/containerd/blob/main/docs/hosts.md#registry-host-namespace). The stargz snapshotter doesn't send it when fetching layer chunks, so mirrors that depend on it can't serve us. Spegel for example rejects requests without `ns`, making it unusable as a mirror with this snapshotter (spegel-org/spegel#510, and see [this comment](https://github.com/spegel-org/spegel/pull/1387#issuecomment-4711974654) suggesting the fix belongs here).

Append `ns=<refspec.Hostname()>` to the blob URL when the host is a mirror, like containerd does. The check is copied from containerd's unexported [`RegistryHost.isProxy`](https://github.com/containerd/containerd/blob/v2.2.3/core/remotes/docker/registry.go#L85-L92), including the special case for `registry-1.docker.io`.

Requests to the ref's own registry are unchanged. Note that chunk cache keys are derived from the blob URL, so mirror users will refetch cached chunks once after upgrading.

### DIFF
--- a/fs/remote/resolver.go
+++ b/fs/remote/resolver.go
@@ -34,6 +34,7 @@ import (
 	"mime"
 	"mime/multipart"
 	"net/http"
+	"net/url"
 	"path"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


simonepri:
@ktock any chance to get your feedback on this? Is there a reason why stargz doesn't include this?

ktock (approved):
Thanks

PR #2385, opened 2026-09-10
fs/metrics: count failed blob reads by errno

kliukovkin:
### Motivation

When a blob read fails on the data path, the failure surfaces only as an error returned up the FUSE stack. Nothing watching the snapshotter can tell one cause from another: a cache volume that has filled up and is failing writes with `ENOSPC` looks exactly like a registry refusing requests, and neither is visible at all unless someone is reading logs at the time.

### What this adds

A counter over the two error paths of `blob.ReadAt`:

```
stargz_fs_blob_fetch_errors_total{errno}
```

The label carries the errno the failure arrived with, extracted with `errors.As` so that it survives wrapping. It is drawn from a fixed set of 17 errnos; anything outside that set is reported as `other`, and an error carrying no errno as `none`, so a storm of failures cannot blow up metric cardinality.

The counter is registered alongside the existing filesystem metrics, so it follows the same `no_prometheus` setting and needs no new configuration. Incrementing it before registration is harmless, and it only happens on a path that is already returning an error, so there is no cost on a healthy read.

No new dependencies.

### How this was verified

- `go build ./...`, `go vet ./...`, `gofmt -l` clean.
- `go test -race ./fs/... ./cache/...` passes.
- New unit tests cover the errno mapping (wrapped errno, bare errno, an errno outside the reported set, an error with no errno) and the counter itself, including the exported metric name and label.

### DIFF
--- a/fs/metrics/common/metrics.go
+++ b/fs/metrics/common/metrics.go
@@ -18,7 +18,9 @@ package commonmetrics
 
 import (
 	"context"
+	"errors"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/containerd/log"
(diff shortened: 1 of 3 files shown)


kliukovkin:
@ktock gentle ping — this is a small, behavior-neutral change (a Prometheus counter over the two existing blob-read error paths, bounded label cardinality). It came out of a measurement study where read failures under cache-volume pressure were invisible in metrics; #2387 is a related finding. Happy to adjust the label set or naming if you'd prefer a different shape. Thanks!

PR #2389, opened 2026-09-17
script/config: set KillMode=process on the stargz-snapshotter unit

### LINKED ISSUE #2387: Shipped systemd unit silently defeats fuse_manager: systemctl restart still unmounts all lazy mounts
## Summary

The FUSE manager (`[fuse_manager] enable = true`) is designed to keep mountpoints and running containers alive across snapshotter restarts. It does cure the permanent-`ESTALE` failure that a plain daemon restart otherwise inflicts on running pods — but only when the snapshotter runs under `KillMode=process`. The systemd unit shipped in this repository (`script/config/etc/systemd/system/stargz-snapshotter.service`) does not set `KillMode`, so systemd's default (`control-group`) applies, and a routine `systemctl restart stargz-snapshotter` SIGTERMs the detached fuse-manager process along with the main daemon. The manager's SIGTERM path unmounts everything (same handler as SIGINT, `fusemanager/fusemanager.go` shutdown path), so with the shipped unit, enabling `fuse_manager` changes nothing: mounts held by running containers still die on every service restart.

Root cause: `fusemanager.go` detaches the manager with `SysProcAttr{Setpgid: true}` — a new *process group*, but not a new *cgroup*. systemd kills by cgroup.

## Measurements

Environment: stargz-snapshotter v0.18.2 (also reproduced building main @ 624678b4), containerd 2.3.1, kind/Kubernetes v1.36.1, Ubuntu 24.04, snapshotter under systemd inside the kind node using the shipped unit verbatim. Workload: a Running pod serving a 140 GB-class eStargz image; after each restart action we attempt reads through the existing lazy mount (8 warm chunks) and a fresh path (4 chunks).

| trial | fuse_manager | KillMode | action | fm PID pre→post | reads after | verdict |
|---|---|---|---|---|---|---|
| T0 | false | default | systemctl restart | — | ESTALE×8 / ENOTCONN | mount broken |
| T1 | true | default (as shipped) | systemctl restart | 11120→11551 | ESTALE×8 / ENOTCONN | mount broken |
| T2 | true | default (as shipped) | systemctl restart | 12014→12432 | ESTALE×8 / ENOTCONN | mount broken |
| T3 | true | process | systemctl restart | 12949→12949 | all clean | mount survived |
| T4 | true | process | systemctl restart | 13789→13789 | all clean | mount survived |
| T5 | true | process | kill -9 (grpc only) | 14584→14584 | all clean | mount survived |
| T6 | true | process | kill -9 (grpc + manager) | 15441→none | ESTALE×8 / ENOTCONN | mount broken (expected) |

With the shipped unit (T1/T2) the fuse-manager PID changes across the restart — systemd killed and the service restart re-spawned it — and the outcome is identical to the fuse_manager=false control (T0). With `KillMode=process` (T3–T5) the manager PID survives and so do the mounts, including across a `kill -9` of the main daemon (T5). All seven outcomes matched predictions pre-registered from source reading before the rig existed. Full per-trial evidence bundles (journal, mounts table, errno censuses, pod events) are available and can be attached.

## Suggested fix

One line in the shipped unit:

```
[Service]
KillMode=process
```

plus a sentence in `docs/overview.md` warning that fuse_manager under systemd requires it (and that `KillMode=control-group`, the default, silently voids the feature). Happy to send the PR.

## Notes

- A related but separate observation from the same runs: with the manager working (KillMode=process), each *graceful* restart duplicates blob-cache contents in place (~18 GB per restart for this image; cumulative, 10 GB → 78 GB over five restarts, with a flat fm-off control). `docs/overview.md` currently describes cache duplication as a consequence of *unexpected* restarts only. I will file that separately with data.
- Context: these measurements come from a study of lazy-pulling failure semantics for model serving ([arXiv:2608.19412](https://arxiv.org/abs/2608.19412)); the restart matrix above was run specifically to check the fuse_manager behavior before reporting it here.

/cc happy to provide the full harness and raw evidence.

pujitha24:
Motivation:
The fuse_manager feature keeps FUSE mountpoints alive across snapshotter
restarts by detaching the fuse manager into its own process
(fusemanager.go uses SysProcAttr{Setpgid: true}, i.e. a new process
group, but not a new cgroup). The shipped systemd unit at
script/config/etc/systemd/system/stargz-snapshotter.service does not set
KillMode, so systemd's default of KillMode=control-group applies: on
`systemctl restart`, systemd signals every process in the service's
cgroup, including the detached fuse manager. The fuse manager's signal
handler treats SIGTERM the same as SIGINT (both trigger unmounting all
snapshots), so with the shipped unit a routine `systemctl restart`
unmounts everything anyway, silently defeating the purpose of enabling
fuse_manager.

Approach:
Set KillMode=process in the [Service] section so systemd only signals
the main containerd-stargz-grpc process on stop/restart, leaving the
detached fuse manager (and the mounts it serves) untouched. Also add a
short note to docs/overview.md's Fuse Manager section pointing this out,
since fuse_manager under systemd silently does nothing useful without
this setting.

Validation:
This is a systemd unit/doc change with no Go code touched, so go
build/test are not meaningfully affected by it. The root cause was
confirmed by reading fusemanager/fusemanager.go, where the shutdown
select loop registers both unix.SIGINT and unix.SIGTERM on the same
channel and runs the same unmounting shutdown path for either signal.
KillMode=process is a standard, documented systemd directive
(systemd.kill(5)) whose semantics (signal only the tracked main PID,
not the whole cgroup) directly address this: it does not affect
Type=notify readiness tracking, and the daemon has no other detached
children in its process tree that would be left orphaned by not doing
a control-group kill. This exact fix (KillMode=process, versus the
shipped unit's default) was already validated live by the issue
reporter across multiple systemctl restart trials on a kind/k8s node
(mounts consistently survived with KillMode=process, and consistently
broke without it, using containerd-stargz-grpc built from this repo's
main branch). No live systemd/kind rig was run in this environment to
independently reproduce that; this disclosure covers that gap.

Report: https://github.com/containerd/stargz-snapshotter/issues/2387
Signed-off-by: Pujitha Paladugu <10557236+pujitha24@users.noreply.github.com>
Assisted-by: claude-sonnet-5 (via Claude Code)

Fixes #2387


docs/overview.md
```diff
@@ -119,6 +119,13 @@ address = "/run/containerd-stargz-grpc/fuse-manager.sock"
 path = "/usr/local/bin/stargz-fuse-manager"
 ```
 
+If you run the snapshotter under systemd, the unit must set `KillMode=process` (as the
+shipped [`stargz-snapshotter.service`](../script/config/etc/systemd/system/stargz-snapshotter.service)
+does). With the default `KillMode=control-group`, systemd signals every process in the
+service's control group on `stop`/`restart`, including the detached fuse manager process,
+which unmounts all snapshots just like a plain snapshotter restart would. This defeats the
+purpose of enabling `fuse_manager`.
```
kliukovkin:
```suggestion
> With systemd's default `KillMode=control-group`, `systemctl restart`
> also SIGTERMs the detached FUSE manager, which unmounts every snapshot.
> The shipped unit sets `KillMode=process` so only `containerd-stargz-grpc`
> is signalled; if you use your own unit, set it too. Note that
> `systemctl stop` then leaves the FUSE manager running — use the SIGINT
> procedure below for a full shutdown.
```

kliukovkin (commented):
Thanks for picking this up! For the record, the measurements from #2387
apply directly to this change: five `systemctl restart` trials each way on
a kind node with a reader pod running — default KillMode: 0/5 mounts
survived (8/8 reads ESTALE); `KillMode=process`: 5/5 survived, 0 read
errors. Raw logs are linked from the issue.

Two small suggestions on the docs note, feel free to ignore:

1. It might be more discoverable under `### FUSE manager mode is enabled`,
   next to the SIGINT shutdown procedure, since that's where the
   restart/stop behaviour is described.
2. Might be worth one sentence that with `KillMode=process` a
   `systemctl stop` also leaves `stargz-fuse-manager` running, so a full
   shutdown still needs the SIGINT procedure.

Possible wording, if useful:

> With systemd's default `KillMode=control-group`, `systemctl restart`
> also SIGTERMs the detached FUSE manager, which unmounts every snapshot.
> The shipped unit sets `KillMode=process` so only `containerd-stargz-grpc`
> is signalled; if you use your own unit, set it too. Note that
> `systemctl stop` then leaves the FUSE manager running — use the SIGINT
> procedure below for a full shutdown.

